PR #165, opened 2020-04-10
Explain how to configure yarn

Plecra:
The packages documentation is still having growing pains, and the syntax for .npmrc is hidden away quite well. This should make getting involved easier for contributors

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Stremio is a modern media center that's a one-stop solution for your video enter
 
 #### Prerequisites:
 * Node.js 10 or higher
-* yarn __configured for use with GitHub Packages__
+* yarn [__configured for use with GitHub Packages__](#authenticating-with-the-github-package-registry)
 
 ### Install dependencies:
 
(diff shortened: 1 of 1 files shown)


nklhtv:
Thank you for the PR, but there is no place for such a detailed explanation in every repository that uses gh packages. A single link to the github documentation is more than enough for everyone who cant search for it.

nklhtv (changes requested):
Thank you for the PR, but there is no place for such a detailed explanation in every repository that uses gh packages. A single link to the github documentation is more than enough for everyone who cant search for it.

Ivshti:
@Plecra I personally like the explanation but we're pushing for a way to get rid of needing this config altogether, so that this repo works with a simple `npm install`, so unfortunately we'll close this and perhaps re-open if we don't manage to make it not require config

davidlondono:
wow really, I didn't know what was those Github Packages config, just found this merge useful and now the devs don't want it?
well you should consider to merge this or other links so you have a community to help you out

Ivshti:
As I said, the intention is to make the GitHub Packages config **unneeded** as soon as possible, which is much better than having to explain it

PR #169, opened 2020-04-21
streams resources: check for video streams added

swetlasg:


nklhtv (changes requested):
This feature will remain part of stremio-core

PR #175, opened 2020-04-28
invalid link urls not rendered

swetlasg:


nklhtv (changes requested):
only half of the linksGroups code is rewritten with Map. the rest of it will just not work.

nklhtv (approved):

Issue #168, opened 2020-04-21
change streaming server url


nklhtv:
add input field in settings that changes the ss url

pierpuccini:
Hey mate! I was wondering how can I contribute by attempting to solve the issue?

PR #178, opened 2020-05-29
change streaming server url

tymmesyde:
#168


nklhtv (changes requested):
Thank you for the PR!
Although its working and the code is great i don't think it provides good UX. It's not clear at all that user must hit enter key after changeing the url.
What i am proposing is to change the url input to a button that opens modal dialog with input field and submit button. 
Check "Add addon" button in addons route. It has the exact same behavior i've described.

tymmesyde:
Yeah, it wasn't ideal..
I think I managed to add your proposal, but I'm not sure about my implementation.
Maybe a dedicated modal dialog like `AddonDetailsModal` would be nice ? Since `addAddonModal` and `editServerUrlModal` _(the one I added)_ are similar.

pedroslopez:
Hi! What's the progress on this? I'd really love this feature

EDIT: After testing this, it seems to work fine. Only thing I would change is the text on the button to save the new streaming server url. At the moment it says "Edit", but I would change it to something like "Save".

src/routes/Settings/Settings.js
```diff
@@ -21,6 +21,42 @@ const Settings = () => {
...
+                action: 'Ctx',
+                args: {
+                    action: 'UpdateSettings',
+                    args: {
+                        ...profile.settings,
+                        streaming_server_url: editServerUrlInputRef.current.value
+                    }
+                }
+            });
+            if (typeof closeEditServerUrlModal === 'function') {
```
nklhtv: no need to check if that is a function here

src/routes/Settings/Settings.js
```diff
@@ -21,6 +21,42 @@ const Settings = () => {
...
+        return [
+            {
+                className: styles['cancel-button'],
+                label: 'Cancel',
+                props: {
+                    onClick: closeEditServerUrlModal
+                }
+            },
+            {
+                label: 'Edit',
```
nklhtv: That sould be named `Submit`

src/routes/Settings/styles.less
```diff
@@ -275,9 +280,72 @@
                                 }
                             }
                         }
+
+                        &.edit-container {
+                            padding: 0;
+
+                            .label {
+                                overflow: hidden;
```
nklhtv: everything is `overflow: hidden` by default

src/routes/Settings/Settings.js
```diff
@@ -354,10 +391,11 @@ const Settings = () => {
                             <div className={styles['option-name-container']}>
                                 <div className={styles['label']}>Url</div>
                             </div>
-                            <div className={classnames(styles['option-input-container'], styles['info-container'], styles['selectable'])}>
-                                <div className={styles['label']}>
-                                    {profile.settings.streaming_server_url}
-                                </div>
+                            <div className={classnames(styles['option-input-container'], styles['edit-container'])}>
+                                <div className={styles['label']}>{profile.settings.streaming_server_url}</div>
```
nklhtv: add title attribute

src/routes/Settings/styles.less
```diff
@@ -11,6 +11,11 @@
     multiselect-label: label;
 }
 
+:import('~stremio/common/ModalDialog/styles.less') {
+    addon-modal-content: modal-dialog-content;
```
nklhtv: that should be named streaming-server-modal-content

src/routes/Settings/Settings.js
```diff
@@ -354,10 +391,11 @@ const Settings = () => {
...
+                                <Button className={classnames(styles['option-input-container'], styles['button-container'])} title={'Edit'} onClick={openEditServerUrlModal}>
```
nklhtv: this is not a good place for reuse settings grid class names
use `configure-button-container` class name and put all of its styles there: flex, display, align-items, justify-content, background-color, width&height: 3.5rem etc

src/routes/Settings/styles.less
```diff
@@ -275,9 +280,72 @@
...
+
+                            .label {
+                                overflow: hidden;
+                                white-space: pre;
+                                text-overflow: ellipsis;
+                                padding: 0 1rem;
+                            }
+
+                            .button-container {
+                                flex: auto;
```
nklhtv: use `flex: none;` so it could remain square

nklhtv (changes requested):
I liked the changes in useProfileSettingsInputs.Could you bring them back? You could name the generic input like `serverUrlInput` with `value` and `onChange(value)` props, so in `editServerUrlOnSubmit` you will call `serverUrlInput.onChange(editServerUrlInputRef.current.value)` and use `serverUrlInput.value` in the JSX code

swetlasg (approved):

nklhtv (approved):

nklhtv:
Fixes #168

PR #201, opened 2020-12-11
Force update dependencies

jimman2003:


nklhtv:
Webpack has already been updated to v5
Updating react to v17 will be considered at a later stage. We have to make sure the breaking changes does not affect the project.

PR #209, opened 2021-02-20
Settings for VSCode

seahindeniz:
Hi,

I'm not sure if anyone uses VSCode as an editor, but this PR adds some abilities which is a bit good for development flow.
I have added some TS related settings and VSCode settings to the project. Furthermore, I've seen some misspelled words and this requires using an extension to avoid that as well so in this PR, after running the VSCode editor, it will pop up a message referring to installation of recommended extensions.

Before the clickable require path:

https://user-images.githubusercontent.com/5789670/108608744-09d5bf80-73da-11eb-91e9-e9c895603228.mp4

After:

https://user-images.githubusercontent.com/5789670/108608729-e743a680-73d9-11eb-9bf5-709ea73645b9.mp4

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,6 @@
+{
+  "recommendations": [
+    "dbaeumer.vscode-eslint",
+    "streetsidesoftware.code-spell-checker"
+  ]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+  "cSpell.words": [
+    "stremio"
+  ]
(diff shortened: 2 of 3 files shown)


nklhtv:
Everyone in the team is using different development environment. Such configurations should be stored in your local clone. If the config files bother you when you commit to git add their paths to gitignore and send PR.

seahindeniz:
What a nice way to reject config files that are not affecting the project's workflow :)
Thanks for the suggesting!

thewickedkarma:
Who said that nobody uses VSCode?

PR #212, opened 2021-02-26
Remove excessive spacing element

seahindeniz:
No visual change, just removal of element

Before:
image

After:
image


nklhtv:
Those changes only works if detail page does not render [vertical nav bar](https://github.com/Stremio/stremio-web/blob/development/src/routes/MetaDetails/MetaDetails.js#L75) which only happens if there are some meta extentions received from addons.

PR #225, opened 2021-06-11
Add loading effect on placeholders

### LINKED ISSUE #206: Show a loading indicator when waiting from a response from an addon
**Description**
We should show some loading indicator when we are waiting for the response from an addon at the moment all what we see is vie with a lot of mocks
**Screenshots**
image
image

tymmesyde:
Related to #206 

New `ShimerEffect` component  used in child divs of placeholders to created a loading/shimmer effect.

stremio-web-loading-final-optimized


nklhtv (changes requested):
It looks nice!
1. package-lock.json should not be modified.
2. Elements added at different times are out of sync. That should be fixable. [This](https://stackoverflow.com/questions/17721269/sync-css-keyframe-color-animations) question is stackoverflow seems related. They suggest using [this](https://github.com/bealearts/css-animation-sync) library. Perhaps we could try it too. The ideal way it not to use a 3rd party library for this but if its not that simple maybe we should.
sync-animation

tymmesyde:
Yes sorry for the `package-lock.json`
I will try to find a way to make this happen !

Issue #228, opened 2021-06-14
When the Stremio Server is missing, open torrent results as typical magnet URIs instead


jaruba:
As per title, when the Stremio Server is not active, we should open torrent results as typical magnet URIs that should be handled by the locally installed torrent client.

nklhtv:
done

PR #230, opened 2021-06-15
Fallback to magnet uri when server is unavailable

tymmesyde:
Related to #228 
It should open a magnet uri in a new tab when the server is unavailable.
It also happens to handle externalUrl.
Hope I did it the right way!


nklhtv (changes requested):
We should use rust for all kind of data manipulations like this.
Modifications should be done in [StreamDeepLinks](https://github.com/Stremio/stremio-core-web/blob/development/src/model/deep_links.rs#L166) like so:
1. declare one more field `external` of type String
2. changes in both `impl From<...> for StreamDeepLinks {  ... }` should be done. Add streaming_server: &StreamingServer argument to both `from` functions and assign correct values based on the state of the streaming server.
3. changes in js are still required to use the correct default `href` and the correct `target` attribute

nklhtv:
i will close this as its moved to the rust layer and the implementation is already in the development branch

Issue #207, opened 2021-01-22
Set focus on add addon text field


TRtomasz:
**Description**
When we click "add addon" button focus should be set to the text field, at the moment it is not 
**Screenshots**
![image](https://user-images.githubusercontent.com/40594253/105492509-f25ac800-5caf-11eb-983f-68c964bb86bb.png)

nklhtv:
Still we have to implement it for all other modals with text input, like streaming server url modal

vczb:
hi @nklhtv I just send a PR with this implementation

PR #231, opened 2021-06-17
Autofocus Add addon modal input

tymmesyde:
Related to #207

### DIFF
--- a/src/routes/Addons/Addons.js
+++ b/src/routes/Addons/Addons.js
@@ -182,7 +182,7 @@ const Addons = ({ urlParams, queryParams }) => {
                             className={styles['addon-url-input']}
                             type={'text'}
                             placeholder={'Paste addon URL'}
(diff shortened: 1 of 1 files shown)


src/routes/Addons/Addons.js
```diff
@@ -182,6 +182,7 @@ const Addons = ({ urlParams, queryParams }) => {
                             className={styles['addon-url-input']}
                             type={'text'}
                             placeholder={'Paste addon URL'}
+                            autoFocus
```
nklhtv: its a minor thing but could you change that to `autoFocus={true}`

nklhtv (changes requested):

src/routes/Addons/Addons.js
tymmesyde: Sure !

nklhtv (approved):

Issue #227, opened 2021-06-14
When playback fails, we should show a message with a button to download an M3U playlist with the stream URL


jaruba:
If playback fails in the web player (if the Stremio Server is not running in order to transcode the video) we should show a message that the video cannot be played in the web player and a button to "download playlist".

The download playlist button should generate an M3U playlist and force download, all local video players should already support M3U playlists, so just opening the downloaded file should start the stream with the local video player.

I prepared some example code that generates an M3U playlist with JS, includes the stream title and stream URL:
```
<a id="downloadlink" href="" download="playlist.m3u">m3u file</a>
<script>
	const streamUrl = 'https://test-videos.co.uk/vids/bigbuckbunny/mp4/h264/1080/Big_Buck_Bunny_1080_10s_1MB.mp4'
	const streamName = 'Big Buck Bunny'
	const playlist = '#EXTM3U\n\n#EXTINF:0,' + streamName + '\n' + streamUrl
	document.getElementById('downloadlink').href = 'data:application/octet-stream;charset=utf-8;base64,' + window.btoa(playlist)
</script>
```

We can also expand on this method by adding the subtitle URL too, but this is only supported by VLC (the other video players would ignore this feature). Here is an example of a M3U file that supports passing a subtitle too:
```
#EXTINF:-1,Abenobashi ED - Anata No Kokoro Ni (In Your Heart)
#EXTVLCOPT:sub-file=/p/Documents/mus/lyrics/Abenobashi ED - Anata No Kokoro Ni (In Your Heart).srt
#EXTVLCOPT:subsdec-encoding=UTF-8
Abenobashi ED - Anata No Kokoro Ni (In Your Heart).flv
```

nklhtv:
done

PR #232, opened 2021-06-19
Allow to download a m3u playlist in case of a playback error

tymmesyde:
Related to #227

### DIFF
--- a/src/routes/Player/styles.less
+++ b/src/routes/Player/styles.less
@@ -46,7 +46,7 @@ html:not(.active-slider-within) {
             background-color: @color-background-dark5;
 
             .error-label {
-                flex: none;
+                flex: auto;
                 padding: 0 8rem;
                 max-height: 4.8em;
                 font-size: 2rem;
@@ -56,7 +56,7 @@ html:not(.active-slider-within) {
 
             .error-details {
                 display: flex;
-                flex: none;
+                flex: auto;
                 font-size: 1.5rem;
                 margin-top: 1.5rem;
                 color: @color-surface-light5-90;
--- a/src/routes/Player/styles.less
+++ b/src/routes/Player/styles.less
@@ -36,6 +36,7 @@ html:not(.active-slider-within) {
         left: 0;
         right: 0;
         bottom: 0;
+        z-index: 0;
 
         &.error-layer {
             display: flex;
@@ -85,7 +86,6 @@ html:not(.active-slider-within) {
         }
 
         &.nav-bar-layer {
-            z-index: 2;
             bottom: initial;
             background: transparent;
             overflow: visible;
@@ -116,7 +116,6 @@ html:not(.active-slider-within) {
         }
 
         &.control-bar-layer {
-            z-index: 2;
             top: initial;
             overflow: visible;
 
@@ -132,7 +131,6 @@ html:not(.active-slider-within) {
         }
 
         &.menu-layer {
-            z-index: 2;
             top: initial;
             left: initial;
             right: 2rem;
--- a/src/routes/Player/styles.less
+++ b/src/routes/Player/styles.less
@@ -36,13 +36,14 @@ html:not(.active-slider-within) {
         left: 0;
         right: 0;
         bottom: 0;
-        z-index: 0;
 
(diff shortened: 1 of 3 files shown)


src/routes/Player/usePlayer.js
```diff
@@ -110,7 +110,19 @@ const usePlayer = (urlParams) => {
         }, 'player');
     }, []);
     const player = useModelState({ model: 'player', action, init, map });
-    return [player, updateLibraryItemState, pushToLibrary];
+    const playlist = React.useMemo(() => {
```
nklhtv: this should be moved to player.js code. later we can extract it into separate file.

src/routes/Player/styles.less
```diff
@@ -36,13 +36,14 @@ html:not(.active-slider-within) {
         left: 0;
         right: 0;
         bottom: 0;
-        z-index: 0;
 
         &.error-layer {
+            z-index: 1;
```
nklhtv: whats the reason of those changes in the zindex?

src/routes/Player/usePlayer.js
```diff
@@ -110,7 +110,19 @@ const usePlayer = (urlParams) => {
...
+        if (player.selected === null || typeof player.selected.stream.url !== 'string') {
+            return null;
+        }
+
+        const playlist = `#EXTM3U\n\n#EXTINF:0,${player.title}\n${player.selected.stream.url}`;
```
nklhtv: player.title must be wrapped in encodeURIComponent
player.selected.stream.url must be wrapped in encodeURI
otherwise if there are some specific characters btoa will throw an exception

src/routes/Player/styles.less
```diff
@@ -53,9 +54,39 @@ html:not(.active-slider-within) {
                 color: @color-surface-light5-90;
                 text-align: center;
             }
+
+            .error-details {
+                gap: 0.8rem;
```
nklhtv: allways declare a flex property explicitly

src/routes/Player/styles.less
```diff
@@ -36,13 +36,14 @@ html:not(.active-slider-within) {
...
-        z-index: 0;
 
         &.error-layer {
+            z-index: 1;
             display: flex;
-            flex-direction: row;
+            flex-direction: column;
             align-items: center;
             justify-content: center;
+            gap: 1.5rem;
```
nklhtv: could this be done my using margin on the child elements?

nklhtv (changes requested):

src/routes/Player/styles.less
tymmesyde: I believe it's because the video-layer happens to be on top of the error-layer, preventing the button to be clicked

src/routes/Player/styles.less
nklhtv: i meant to allways declare a prop like `flex: XXX` explicitly when parent has `display: flex`

src/routes/Player/styles.less
nklhtv: i see the issue now.
the correct way to solve this is by fixing the markup hierarchy.
revert the z-index changed.
move the `{error !== null ? </> : null}` block right after the layer with onVideoClick/onVideoDoubleClick and just before the layer that is shown when `subtitlesMenuOpen || infoMenuOpen`

src/routes/Player/styles.less
```diff
@@ -53,9 +53,41 @@ html:not(.active-slider-within) {
...
+                font-size: 1.5rem;
+                margin-top: 1.5rem;
+                color: @color-surface-light5-90;
+
+                .error-details-button {
+                    flex: none;
+                    display: flex;
+                    flex-direction: row;
+                    align-items: center;
+                    gap: 1rem;
```
nklhtv: there is still usage of `gap` here

src/routes/Player/styles.less
nklhtv: there is still usage of `gap` here

nklhtv (changes requested):

nklhtv (approved):

Issue #217, opened 2021-03-23
MetaPreview Show button


swetlasg:
add the Show button in meta preview when on discover page

abhakta-47:
Is any body working on this?
If not i would like to contribute.

PR #235, opened 2021-06-29
show button added to MetaPreview

abhakta-47:
In discover page, show button added to MetaPreview, which opens MetaDetails    #217 .

### DIFF
--- a/src/common/MetaPreview/MetaPreview.js
+++ b/src/common/MetaPreview/MetaPreview.js
@@ -210,16 +210,19 @@ const MetaPreview = ({ className, compact, name, logo, background, runtime, rele
                         null
                 }
                 {
-                    <ActionButton
-                        className={styles['action-button']}
-                        icon={'ic_play'}
-                        label={'Show'}
-                        tabIndex={compact ? -1 : 0}
-                        href={showHref}
-                    />
+                    typeof showHref === 'string' && compact ?
+                        <ActionButton
+                            className={styles['action-button']}
+                            icon={'ic_play'}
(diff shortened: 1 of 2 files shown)


src/common/MetaPreview/MetaPreview.js
```diff
@@ -194,6 +209,15 @@ const MetaPreview = ({ className, compact, name, logo, background, runtime, rele
                         :
                         null
                 }
+                {
+                    <ActionButton
```
nklhtv: you need to check for `typeof showHref === 'string' && compact` to make sure this button is rendered only if there is a valid href and user is in discover

src/common/MetaPreview/MetaPreview.js
```diff
@@ -194,6 +209,15 @@ const MetaPreview = ({ className, compact, name, logo, background, runtime, rele
...
+                        className={styles['action-button']}
+                        icon={'ic_play'}
+                        label={'Show'}
+                        tabIndex={compact ? -1 : 0}
+                        href={showHref}
+                    />
+                }
                 {
                     linksGroups.has(CONSTANTS.SHARE_LINK_CATEGORY) ?
```
nklhtv: the discover ui doesnt look good when we have 4 buttons here. add a check for `!compact` here so that share button is allways hidden in discover

nklhtv (changes requested):

nklhtv (approved):

nklhtv:
Thank you @abhakta-47. I am merging this.

Issue #216, opened 2021-03-23
MetaDetails videos when on streams


swetlasg:
show videos when on streams in meta details page

nklhtv:
Show VideosList next to StreamsList in the MetaDetails UI when there is no `behaviorHints.defaultVideoId` in the meta item

anmol210202:
I wanna work on this issue . But i need more details . Can  anyone explain ???

PR #236, opened 2021-06-30
urlparam exception for movies fixed

abhakta-47:
issue #216 
router and urlparam are written such that,
urls are marched against
`/metadetails/<type>/<id>/<videoid>` this type of url,
for series, where there are multiple episodes per show, show is identified by `id` and individual episode by `videoid`
but for movies there is no episode, movie itself is video, so `id` and `videoid` is one and same,
so, `/metadetails/movie/<id>/` should give expected result, but urlparamsforpath set `videoid=null`,
that is fixed

### DIFF
--- a/src/router/Router/urlParamsForPath.js
+++ b/src/router/Router/urlParamsForPath.js
@@ -2,13 +2,17 @@
 
 const urlParamsForPath = (routeConfig, path) => {
     const matches = path.match(routeConfig.regexp);
+
+    if (matches[1] === 'movie' && typeof matches[3] != 'string') {
+        matches[3] = matches[2]
+    }
+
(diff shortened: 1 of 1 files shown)


jaruba:
The presumption that the `meta.id` is the same as `video.id` for movies is not correct.

See the addon SDK docs here: https://github.com/Stremio/stremio-addon-sdk/blob/master/docs/api/responses/meta.md

More specifically `meta.behaviorHints.defaultVideoId`:
> string, set to a Video Object id in order to open the Detail page directly to that video's streams

One such addon that has different values for `meta.id` and `video.id` is "The Movie Database Addon".

nklhtv:
`videoId` is a required url parameter because stremio-web is type agnostic. The issue #216 liked here is not related to url parameters. I will add a bit more description there.

PR #239, opened 2021-07-19
Seek time settings and player

tymmesyde:
Add a new setting option in Settings to allow user to change the default value of the seek duration keys in Player.

### DIFF
--- a/src/routes/Player/Player.js
+++ b/src/routes/Player/Player.js
@@ -344,7 +344,7 @@ const Player = ({ urlParams, queryParams }) => {
                 }
                 case 'ArrowRight': {
                     if (!subtitlesMenuOpen && !infoMenuOpen && videoState.time !== null) {
-                        let seekTimeDuration = settings.seekTimeDuration * 1000;
+                        let seekTimeDuration = settings.seekTimeDuration;
 
                         if (event.shiftKey === true) {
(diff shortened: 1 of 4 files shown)


nklhtv (changes requested):
`settings.seekTimeDuration` must be added in the effect hook for onKeyDown listener

src/routes/Player/Player.js
```diff
@@ -344,14 +344,26 @@ const Player = ({ urlParams, queryParams }) => {
...
                     }
 
                     break;
                 }
                 case 'ArrowLeft': {
                     if (!subtitlesMenuOpen && !infoMenuOpen && videoState.time !== null) {
-                        onSeekRequested(videoState.time - 15000);
+                        let seekTimeDuration = settings.seekTimeDuration;
+
+                        if (event.shiftKey === true) {
```
nklhtv: expressions like `if (x === true)` looks wierd. better use `if (x)`

nklhtv: or maybe even better like:
```
const seekTimeMultiplier = event.shiftKey ? 3 : 1;
onSeekRequested(....);
```

nklhtv (changes requested):

nklhtv (approved):

Issue #248, opened 2021-10-23
Sorting by year still randomly assorts the movie all throughout the year


Jinkerz94:
When selecting _Discover > Movies > By Year_ or _Discover > Series > By year_ the movies/series are randomly assorted all over the year specified. To look for new content, the user is required to examine the entire year's section over and over whenever he/she is looking for new content.  (e.g. if I select 2020, all the content from 2020 appears. But if new content from 2020 was added, it could appear anywhere in the entire 2020 section, and you will need to scan the entire section just to notice it).

It would be beneficial to make new movies/series appear at the top of the list, or at least sort it by month of the year, to avoid having to look through all the content one has either added to the watchlist or simply want to avoid watching.

nklhtv:
this was an issue with the cinemeta addon and it is now fixed

PR #250, opened 2021-10-27
#248 modify useDiscover to sort by date

abhakta-47:
#248 resolve by modifying useDiscover.js to sort catalog items by date when 'By Year' is selected


nklhtv:
Thank you for the PR!
The order of the items in the catalog must be handled by the addon, not in the UI.
I've already pruposed changes in the Cinemeta addon related to this issue.

PR #260, opened 2022-02-04
Fix Broken Image Icon Showing for Backdrops on MetaDetails Page

jaruba:
Because the backdrop was not set as a `background-image`, a broken image icon could be seen when the backdrop was missing:
Screenshot 2022-01-26 at 19 06 36


nklhtv (changes requested):
renderFallback api should be used instead

nklhtv:
implemented in #development

PR #262, opened 2022-02-04
Fallback to Using Addon Name if Stream Name is Missing

jaruba:
We use `stream.name` for the left side of the stream element, this is fine in most cases but some don't set `stream.name`, in this case we should fallback to the addon's name.

Before:
Screenshot 2022-02-04 at 17 01 05

After:
Screenshot 2022-02-04 at 17 05 33

### DIFF
--- a/src/routes/MetaDetails/StreamsList/Stream/Stream.js
+++ b/src/routes/MetaDetails/StreamsList/Stream/Stream.js
@@ -35,7 +35,7 @@ const Stream = ({ className, addonName, name, description, thumbnail, progress, 
                     </div>
                     :
                     <div className={styles['addon-name-container']} title={name}>
-                        <div className={styles['addon-name']}>{name}</div>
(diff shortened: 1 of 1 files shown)


src/routes/MetaDetails/StreamsList/Stream/Stream.js
```diff
@@ -35,7 +35,7 @@ const Stream = ({ className, addonName, name, description, thumbnail, progress,
                     </div>
                     :
                     <div className={styles['addon-name-container']} title={name}>
-                        <div className={styles['addon-name']}>{name}</div>
+                        <div className={styles['addon-name']}>{name || addonName}</div>
```
nklhtv, line 38: add the addonName fallback in the title attribute above too

nklhtv (changes requested):

nklhtv (approved):

PR #263, opened 2022-02-05
Copy magnet link of the stream to clipboard on play

parz3val:
It is lot difficult to get magnet link if stream fails to load somehow. 
This feature automatically copies magnet link of stream to clipboard in case stream fails to play, so you can download the video with other torrent clients.


jaruba:
@parz3val But if the stream fails it already shows a button with "open external" that points to the magnet link

jaruba:
Closing this due to a lack of activity, if you can motivate this PR later on, feel free to open a discussion and we may re-open the PR.

PR #274, opened 2022-04-06
Add Dockerfile

egikyan:
- Add simple Dockerfile
- Change access of external repo as github deprecated [unsecure git](https://github.blog/2021-09-01-improving-git-protocol-security-github/#no-more-unauthenticated-git)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+# Stremio Node 14.x
+FROM stremio/node-base:fermium
+
+# Meta
+LABEL Description="Stremio Web" Vendor="Smart Code ltd" Version="1.0.0"
+
+# Create app directory
+RUN mkdir -p /var/www/stremio-web
+
+# Install app dependencies
+WORKDIR /var/www/stremio-web
+COPY package.json package-lock.json /var/www/stremio-web
+RUN npm install 
+RUN npm install -g http-server
+
+# Bundle app source
+WORKDIR /var/www/stremio-web
+COPY . /var/www/stremio-web
+
+RUN npm run build 
+
+EXPOSE 8080
+CMD ["http-server", "/var/www/stremio-web/build/", "-p", "8080"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,22 +2,21 @@
 FROM stremio/node-base:fermium
 
 # Meta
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+# Meta
+LABEL Description="Stremio Web" Vendor="Smart Code ltd" Version="1.0.0"
+
+# Create app directory
+RUN mkdir -p /var/www/stremio-web
+
+# Install app dependencies
+WORKDIR /var/www/stremio-web
+COPY package.json package-lock.json /var/www/stremio-web
```
core1024: This step doesn't work, at least with my version of Docker

```
Step 5/12 : COPY package.json package-lock.json /var/www/stremio-web
When using COPY with more than one source file, the destination must be a directory and end with a /
```
 Also it is redundant as latter on you do `COPY . /var/www/stremio-web` that copies all the files. You can just move that line here.

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN npm install -g http-server
+
+# Bundle app source
+WORKDIR /var/www/stremio-web
+COPY . /var/www/stremio-web
+
+RUN npm run build 
+
+EXPOSE 8080
+CMD ["http-server", "/var/www/stremio-web/build/", "-p", "8080"]
```
core1024: Would you also add `-d false` here. Disabling directory listings is always a good idea.

core1024 (changes requested):

Dockerfile
```diff
@@ -0,0 +1,23 @@
+# Stremio Node 14.x
+FROM stremio/node-base:fermium
+
+# Meta
+LABEL Description="Stremio Web" Vendor="Smart Code ltd" Version="1.0.0"
```
nklhtv: should be "Smart Code OOD"

Dockerfile
egikyan: Done

Dockerfile
egikyan: Done

Dockerfile
egikyan: Done

core1024 (approved):

PR #277, opened 2022-04-15
Pass `seriesInfo` to Player

jaruba:
The video player is relying on `seriesInfo` to be passed too (which includes season and episode, if available), this was never implemented in Stremio Web though.

The problem with this PR is that until now we started the video player instantly as we didn't need any metadata to initiate playback, this PR changes things though as we need to wait for the metadata request to finish in order to have `seriesInfo` available, which is needed to start playback.

This should not be an issue though, as it should not be possible to get to the video player page without going through the meta page, so the meta request should be handled by the local browser cache.

Related PRs:
- https://github.com/Stremio/stremio-video/pull/20
- https://github.com/Stremio/enginefs/pull/19

### DIFF
--- a/src/routes/Player/Player.js
+++ b/src/routes/Player/Player.js
@@ -213,12 +213,12 @@ const Player = ({ urlParams, queryParams }) => {
             let seriesInfo = false;
             if ((player.metaItem?.videos || []).length) {
                 const videoId = player.selected?.streamRequest?.path?.id;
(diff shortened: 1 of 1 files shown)


nklhtv:
seriesInfo will be exposed from player model in core

nklhtv:
exposed from stremio-core and sent to video impl in #development

PR #291, opened 2022-07-17
Add meta tags for iOS PWA support

nperez0111:
Add a couple of meta tags that allow "Add to Homescreen" on iOS to: A) Use the correct icon B) not just open the website in Safari, but open as a fullscreen app

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -4,7 +4,10 @@
 <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1">
+    <meta name="apple-mobile-web-app-capable" content="yes">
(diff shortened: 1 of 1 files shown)


nklhtv:
That looks nice!
can you add `apple-mobile-web-app-title` `Stremio` too ?
does `apple-touch-icon` needs to be after `title`? can you move it after `icon`?

nklhtv (changes requested):
That looks nice!
can you add apple-mobile-web-app-title Stremio too ?
does apple-touch-icon needs to be after title? can you move it after icon?

nperez0111:
Done & Done

nklhtv:
@nperez0111 Thank you for the contribution!

nklhtv (approved):

PR #292, opened 2022-07-26
Implement shell IPC and pass it to the player

core1024:


src/routes/Player/Player.js
```diff
@@ -263,6 +263,7 @@ const Player = ({ urlParams, queryParams }) => {
                 }
             }, {
                 chromecastTransport: chromecast.active ? chromecast.transport : null,
+                shell,
```
nklhtv: can you change that to:
`shell: shell.active ? shell.transport : null`
so that we dont send ref to the service

nklhtv (changes requested):

src/routes/Player/Player.js
core1024: Changed as requested

nklhtv (approved):

PR #308, opened 2022-10-28
feat: change font with variable Roboto Flex font

yilmazcabuk:
With the development of the modern web, old static fonts are now being replaced by variable fonts. As we can see in [this](https://material.io/blog/roboto-flex) article from Google, the variable version of the Roboto font appears as Roboto Flex.


src/common/AddonDetailsModal/AddonDetails/styles.less
```diff
@@ -8,7 +8,8 @@
         flex-direction: row;
         align-items: center;
 
-        .logo, .icon {
```
nklhtv, line 11: Those changes must be reverted and sent in a separate PR

src/common/CONSTANTS.js
```diff
@@ -2,7 +2,7 @@
 
 const CHROMECAST_RECEIVER_APP_ID = '1634F54B';
 const SUBTITLES_SIZES = [75, 100, 125, 150, 175, 200, 250];
-const SUBTITLES_FONTS = ['Roboto', 'Arial', 'Halvetica', 'Times New Roman', 'Verdana', 'Courier', 'Lucida Console', 'sans-serif', 'serif', 'monospace'];
+const SUBTITLES_FONTS = ['Roboto Flex', 'Roboto', 'Arial', 'Halvetica', 'Times New Roman', 'Verdana', 'Courier', 'Lucida Console', 'sans-serif', 'serif', 'monospace'];
```
nklhtv, line 5: Should we keep 'Roboto' here or replace it with 'Roboto Flex' instead?

.gitignore
```diff
@@ -3,3 +3,4 @@
 /yarn.lock
 /npm-debug.log
 .DS_Store
+compose-dev.yaml
```
nklhtv, line 6: Why `compose-dev.yaml` is added to .gitignore

nklhtv (changes requested):

src/common/CONSTANTS.js
yilmazcabuk: In my opinion we should keep 'Roboto' as a fallback option.

PR #311, opened 2022-10-31
feat(Player): add options menu

tymmesyde:


src/routes/Player/Player.js
```diff
@@ -554,6 +563,15 @@ const Player = ({ urlParams, queryParams }) => {
                     :
                     null
             }
+            {
+                optionsMenuOpen ?
+                    <OptionsMenu
+                        className={classnames(styles['layer'], styles['menu-layer'])}
+                        stream={videoState.stream}
```
nklhtv: player.selected.stream might be better here, because when the video throws error, videoState.stream will be null

nklhtv (changes requested):
Screen Shot 2022-11-02 at 12 24 52-fullpage
Is that expected UX?

src/routes/Player/Player.js
tymmesyde: Oh okay !

src/routes/Player/OptionsMenu/OptionsMenu.js
```diff
@@ -0,0 +1,97 @@
...
+                        message: 'Stream link was copied to your clipboard',
+                        timeout: 3000
+                    });
+                })
+                .catch((e) => {
+                    console.error(e);
+                    toast.show({
+                        type: 'error',
+                        title: 'Error',
+                        message: 'Failed to copy stream link. Try to enable your browser permission.',
```
nklhtv: can we also unclude the `streamUrl` into the message

src/routes/Player/OptionsMenu/OptionsMenu.js
```diff
@@ -0,0 +1,97 @@
...
+const { Button, useStreamingServer, useToast } = require('stremio/common');
+// const { useServices } = require('stremio/services');
+const styles = require('./styles');
+
+const OptionsMenu = ({ className, stream }) => {
+    const streamingServer = useStreamingServer();
+    // const { core } = useServices();
+    const toast = useToast();
+    const streamUrl = React.useMemo(() => {
+        return stream !== null ?
```
nklhtv, line 15: ytId case must also be handled here.
we kind of handle it already in [external_player](https://github.com/Stremio/stremio-core/blob/development/src/deep_links/mod.rs#L239) but the url case is not suitable for this feature.
I suggest we refactor stremio-core by adding one more field `url` to ExternalPlayerLink which will be the same as href except for [this](https://github.com/Stremio/stremio-core/blob/development/src/deep_links/mod.rs#L31) case, then use deepLinks api here

nklhtv (changes requested):

elpiel (approved):

Issue #296, opened 2022-07-29
Add `cause` argument to the custom errors


nklhtv:
Adding `cause` can help debugging.
For example [this](https://github.com/Stremio/stremio-web/blob/development/src/services/Chromecast/Chromecast.js#L23) Error should look like:
```js
error = new Error('Google Cast API not available', { cause: args });
```

zeerakb1:
hello @nklhtv, can I work on this issue?

zeerakb1:
hello @nklhtv, can I work on this issue?

nklhtv:
@zeerakb1 yes

alfi-j:
Is this issue still open for collaborations?
Just for clarify things, you need that every new error have a **cause** parameter so it can be easily debugged, is that?

PR #320, opened 2022-11-08
Add cause argument to the custom errors #296

zafar-hussain:
Added cause to help in debugging


nklhtv:
Thank you @zafar-hussain 
I really like those changes.
Could you add `cause` to the other errors too?

nklhtv (approved):

PR #329, opened 2022-12-16
Interface language

### LINKED ISSUE #347: Translations and interface language
We need to enable the UI language setting on Web.

There has been some effort to do it in #329 but it's an old PR with conflicts and should be either updated or we should transfer the changes to a new branch.

Related issues:
#108

tymmesyde:


elpiel (changes requested):
- [x] ~it seems that when you open the Settings it does not load the user chosen language in the dropdown. After changing the language, I do not experience this issue anymore. Was the value somehow wrong and the dropdown couldn't select the given language?!~ We haven't been able to reproduce this issue, if it happens again then a bug should be raised.

Screenshot from 2023-03-15 10-48-59

- [x] It would be nice feature to scroll the dropdown to the chosen language if this could be easily achieved with the current dropdown impl #350 

Another thing that I noticed is that Settings and other parts of the application are still not translated but that is part of the `stremio-translations` strings rather than this PR.

src/App/App.js
```diff
@@ -90,6 +92,21 @@ const App = () => {
         };
     }, []);
     React.useEffect(() => {
+        const onCoreEvent = ({ event, args }) => {
+            switch (event) {
+                case 'SettingsUpdated': {
+                    if (args && args.settings && typeof args.settings.interfaceLanguage === 'string') {
+                        i18n.changeLanguage(args.settings.interfaceLanguage);
+                    }
```
sleeyax, line 100: You can use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) to simplify this. All latest browsers except IE already [support](https://caniuse.com/?search=optional%20chain) it and it's included in babel's `@babel/preset-env` preset.

```suggestion
                   const language = args?.settings?.interfaceLanguage;
                    if (typeof language === 'string') {
                        i18n.changeLanguage(language);
                    }
```
The additional const variable assignment in my suggestion is optional, but makes it even more readable imo.

src/App/App.js
```diff
@@ -90,6 +92,21 @@ const App = () => {
...
+                    break;
+                }
+            }
+        };
+        const onCtxState = (state) => {
+            if (state && state.profile && state.profile.settings && typeof state.profile.settings.interfaceLanguage === 'string') {
```
sleeyax, line 106:
```suggestion
            if (typeof state?.profile?.settings?.interfaceLanguage === 'string') {
```

src/App/App.js
```diff
@@ -113,8 +130,16 @@ const App = () => {
         if (services.core.active) {
             onWindowFocus();
             window.addEventListener('focus', onWindowFocus);
+            services.core.transport.on('CoreEvent', onCoreEvent);
+            services.core.transport
+                .getState('ctx')
+                .then(onCtxState)
+                .catch((e) => console.error(e));
```
sleeyax, line 137: Nitpicking but this can be even shorter :) 

```suggestion
                .catch(console.error);
```

src/common/LibItem/LibItem.js
```diff
@@ -4,29 +4,35 @@ const React = require('react');
...
-                    return typeof _id === 'string' && removable;
-            }
-        });
+        return OPTIONS
+            .filter(({ value }) => {
+                switch (value) {
+                    case 'play':
+                        return props.deepLinks && typeof props.deepLinks.player === 'string';
+                    case 'details':
+                        return props.deepLinks && (typeof props.deepLinks.metaDetailsVideos === 'string' || typeof props.deepLinks.metaDetailsStreams === 'string');
```
sleeyax, line 25:
```suggestion
                    case 'play':
                        return typeof props.deepLinks?.player === 'string';
                    case 'details':
                        return (typeof props.deepLinks?.metaDetailsVideos === 'string' || typeof props.deepLinks?.metaDetailsStreams === 'string');
```

src/index.js
```diff
@@ -13,8 +13,26 @@ if (browser?.platform?.type === 'desktop') {
...
+const i18n = require('i18next');
+const { initReactI18next } = require('react-i18next');
+const stremioTranslations = require('stremio-translations');
 const App = require('./App');
 
+const translations = Object.fromEntries(Object.entries(stremioTranslations()).map(([key, value]) => [key, {
+    translation: value
+}]));
+
+i18n
```
sleeyax, line 25: Consider adding [i18next-browser-languagedetector](https://www.npmjs.com/package/i18next-browser-languagedetector) so the webapp tries to select whatever language the browser is set to before falling back to en-US.

src/common/translateOption.js
```diff
@@ -0,0 +1,18 @@
...
+    const translateKeyUppercase = translateKey.toUpperCase();
+    const translateValue = t(translateKey);
+    const translateValueUppercase = t(translateKeyUppercase);
+    if (translateKey !== translateValue) {
+        return translateValue;
+    } else if (translateKeyUppercase !== translateValueUppercase) {
+        return translateValueUppercase;
+    }
+    return option.charAt(0).toUpperCase() + option.slice(1);
+};
```
sleeyax: This function is relatively hard to follow. If I understand what it's doing correctly, I believe it can be heavily simplified as follows:
```suggestion
const translateOption = (option, translateKeyPrefix = '') => {
    const translateKey = `${translateKeyPrefix}${option}`;
    const translateValue = t(translateKey, {defaultValue: t(translateKey.toUpperCase(), {defaultValue: null})});
    return translateValue ?? option.charAt(0).toUpperCase() + option.slice(1);
};
```

sleeyax (dismissed):
I commented some minor suggestions.

sleeyax:
- [ ] It would also be nice if the user could type the language they are looking for as they open the language selector. Just like elpiel mentioned, I'm not sure how hard this would be to add to the current dropdown implementation, but this is definitely something builtin into most browsers native `select` element and could work the same way here.

src/App/App.js
tymmesyde: I did not used optional chaining because i wanted to follow the current way everything is checked in the app
Also i think in js it might be clearer to spot mistakes with this syntax since there are no linting

tymmesyde:
> * [ ]  it seems that when you open the Settings it does not load the user chosen language in the dropdown. After changing the language, I do not experience this issue anymore. Was the value somehow wrong and the dropdown couldn't select the given language?!

I cannot reproduce this issue neither on chrome or firefox as guest or logged in, what steps did you take ?

elpiel:
@tymmesyde I'm not sure how to reproducing this. I tried clearing cache and to login again but I don't see this `afr` value anymore.
Is it possible that my User data was corrupted with a bad value for the language in the first place?

@sleeyax I didn't mean search but for the select to scroll down to the picked choice when opened.
But a search functionality might be a good idea in the long run, though, we don't have that many languages and that's why I think this is low priority.

sleeyax:
> @tymmesyde I'm not sure how to reproducing this. I tried clearing cache and to login again but I don't see this `afr` value anymore.
> Is it possible that my User data was corrupted with a bad value for the language in the first place?
> 
> @sleeyax I didn't mean search but for the select to scroll down to the picked choice when opened.
> But a search functionality might be a good idea in the long run, though, we don't have that many languages and that's why I think this is low priority.

Yes, I know, I just added my own suggestion on top :)

elpiel (dismissed):
LGTM! @tymmesyde Decide if you'd like to tackle PR comments from @sleeyax in this PR or can be done in subsequent PR. At this stage I consider the PR ready.

src/common/translateOption.js
tymmesyde: Much better thank you !

src/App/App.js
elpiel: let's indeed keep it consistent and update it all at once if the time arises

package.json
```diff
@@ -35,8 +36,10 @@
         "react": "18.2.0",
         "react-dom": "18.2.0",
         "react-focus-lock": "2.9.1",
+        "react-i18next": "^12.1.1",
         "react-is": "18.2.0",
         "spatial-navigation-polyfill": "git+https://git@github.com/Stremio/spatial-navigation.git#64871b1422466f5f45d24ebc8bbd315b2ebab6a6",
+        "stremio-translations": "git+https://git@github.com/Stremio/stremio-translations.git#ef047329f5bcb0a8f96008fca05c68b449b34cb1",
```
elpiel, line 42: We can keep translations up-to-date using version to keep track which translations we have in the app, with commit it's not that visible.
- I'll look into releasing versions to `npm` and just use Patch versions and `npm update`.
- ~Could you use a tag version for now?~
```suggestion
        "stremio-translations": "git+https://git@github.com/Stremio/stremio-translations.git#v1.43.16",
```

elpiel (approved):
LGTM - Tested locally and it works.

package.json
tymmesyde: I didn't even know that we started to do tag versioning for stremio-translations
I guess it makes sense to use it if we actually have a regular release cycle, the advantage of using commit hash was that it was quick to update in any app, translations are sparse and doesn't come in bulk most of the time

src/App/App.js
sleeyax: There are other places where optional chaining is used. So an inconsistency already exists. I agree this is out of scope for this PR though, we can pick one style and create a refactoring PR later.

sleeyax (approved):

package.json
elpiel: Yes, correct, we will leave it for now to hashes.
There are inherit complications with releasing versions that we will avoid for now.
I've discussed this both with @TheBeastLT and @jaruba

PR #330, opened 2022-12-22
Feat/data export

tymmesyde:
Waiting for this: https://github.com/Stremio/stremio-core/pull/411


elpiel (dismissed):
I've tested it locally and it works.

PR #344, opened 2023-03-01
fix: redirect to home when user is already logged #328

### LINKED ISSUE #328: Able to access login screen when already logged-in
## How to produce the bug

1. Login to your stremio account
2. go to https://web.stremio.com/#/intro
3. Login as guess

## Actual result
You're able to get to the login screen even though you're already logged in

## Expected result
You should be redirected to the home screen, you're authenticated after all

vczb:
Description: 

If the user is already logged in redirect to home if navitate to login page

---
Implementation details: 

- Add `withCoreSuspender` Context to `Intro` page
- Get the user profile with `useProfile` hook
- If user is authenticated redirect to home

---

Demo video: 

https://user-images.githubusercontent.com/51221635/222294227-531e5b4c-1a3f-47e4-acd9-40e511f93bc2.mp4

---

Close #328

### DIFF
--- a/src/routes/Intro/Intro.js
+++ b/src/routes/Intro/Intro.js
@@ -11,6 +11,8 @@ const CredentialsTextInput = require('./CredentialsTextInput');
 const ConsentCheckbox = require('./ConsentCheckbox');
 const PasswordResetModal = require('./PasswordResetModal');
 const useFacebookToken = require('./useFacebookToken');
+const useProfile = require('stremio/common/useProfile');
+const { withCoreSuspender } = require('stremio/common/CoreSuspender');
 const styles = require('./styles');
 
 const SIGNUP_FORM = 'signup';
@@ -18,6 +20,7 @@ const LOGIN_FORM = 'login';
 
 const Intro = ({ queryParams }) => {
     const { core } = useServices();
+    const profile = useProfile();
     const routeFocused = useRouteFocused();
     const getFacebookToken = useFacebookToken();
     const emailRef = React.useRef(null);
@@ -29,6 +32,7 @@ const Intro = ({ queryParams }) => {
     const errorRef = React.useRef(null);
     const [passwordRestModalOpen, openPasswordRestModal, closePasswordResetModal] = useBinaryState(false);
     const [loaderModalOpen, openLoaderModal, closeLoaderModal] = useBinaryState(false);
+
     const [state, dispatch] = React.useReducer(
         (state, action) => {
             switch (action.type) {
@@ -278,6 +282,12 @@ const Intro = ({ queryParams }) => {
             core.transport.off('CoreEvent', onCoreEvent);
         };
     }, [routeFocused]);
+
+    if(profile?.auth?.key) {
+        window.location = '/';
+        return;
+    }
+
     return (
         <div className={styles['intro-container']}>
             <div className={styles['form-container']}>
@@ -411,4 +421,4 @@ Intro.propTypes = {
     queryParams: PropTypes.instanceOf(URLSearchParams)
 };
 
-module.exports = Intro;
+module.exports = withCoreSuspender(Intro);


src/routes/Intro/Intro.js
```diff
@@ -278,6 +282,12 @@ const Intro = ({ queryParams }) => {
             core.transport.off('CoreEvent', onCoreEvent);
         };
     }, [routeFocused]);
+
+    if(profile?.auth?.key) {
```
nklhtv, line 286: Could you wrap this in useEffect with [] as deps so it gets executed only on mount?

nklhtv (changes requested):

src/routes/Intro/Intro.js
vczb: wrapping it to an `useEffect` the login page will render until the `profile` load, take a look at it:

https://user-images.githubusercontent.com/51221635/222966364-8fd2dcb5-8f2f-484a-9ba8-22ed093beece.mp4

vczb: > Could you wrap this in useEffect with [] as deps so it gets executed only on mount?

hello @nklhtv I think that the `useEffect` is not needed in the conditional render. 

please check the documentation https://beta.reactjs.org/learn/conditional-rendering and give me your thoughts

nklhtv: I agree with you. useEffect here was not a good idea!
Another idea that i have is to add another callback prop to Router like `onRouteChange(route, urlParams, queryParams)` or smtn, which returns boolean and is invoked [here](https://github.com/Stremio/stremio-web/blob/development/src/router/Router/Router.js#L45), just before `setViews()`. Based on the result, the router can decide to change the location to the previous hash (provided in args of the hashchange event, event.oldURL)
The advantage here is that we have access to the old URL, but its a bit complex to implement.
1. you have to wrap the Router in a HOC that has access to useProfile()
2. you have to declare a callback that makes the required checks inside this new component
3. pass that cb to the Router

not too sure about that but i think it should looks like the pseudo-code bellow:

```
const withThatNewHOC = (Component) => {
    return function withThatNewHOC(props) {
        const profile = useProfile();
        const onRouteChange = useCallback((route, urlParams, queryParams) => true)
        return <Component {...props} onRouteChange={onRouteChange} />
    }
}
const StremioRouter = withCoreSuspender(withThatNewHOC(Router))
....
<StremioRouter .... />
```
What do you think? I am open to other suggestions.

vczb: I will take a look and back to you ASAP

vczb: Hey @nklhtv thanks for your suggestions. Infortunatelly it is not too much clear to me, I just update this PR with a WIP pseudo-code implementation as you suggest and I'm not sure how to check if the user is authenticated or not properly

TRtomasz: @vczb  you can use `profile.auth` i think this is what @nklhtv  meant like here auth is null so display stuff for anonymous user
https://github.com/Stremio/stremio-web/blob/d44f7047dcee5f0b2257318ac7153a09bbfe2959/src/common/NavBar/HorizontalNavBar/NavMenu/NavMenuContent.js#L44

src/App/App.js
```diff
@@ -12,6 +12,7 @@ const DeepLinkHandler = require('./DeepLinkHandler');
 const ErrorDialog = require('./ErrorDialog');
 const routerViewsConfig = require('./routerViewsConfig');
 const styles = require('./styles');
+const StremioRouter = require('stremio/router/Router/StremioRouter');
```
tymmesyde: Here it should be something like this according to Nikola's comment
```js
const { Router, withAuthCheck } = require('stremio-router');
const { withCoreSuspender } = require('stremio/common');

const StremioRouter = withCoreSuspender(withAuthCheck(Router));
```

src/router/Router/StremioRouter.js
```diff
@@ -0,0 +1,17 @@
+const React = require('react');
+const useProfile = require('stremio/common/useProfile');
+const Router = require('./Router');
+const { withCoreSuspender } = require('stremio/common');
```
tymmesyde: Remove
```js
const Router = require('./Router');
const { withCoreSuspender } = require('stremio/common');
```

src/router/Router/StremioRouter.js
```diff
@@ -0,0 +1,17 @@
...
+        const profile = useProfile();
+        // eslint-disable-next-line no-unused-vars
+        const onRouteChange = React.useCallback((route, urlParams, queryParams) => true);
+        return <Component {...props} onRouteChange={onRouteChange} profile={profile} />;
+    };
+};
+
+const StremioRouter = withCoreSuspender(withAuthCheck(Router));
+
+module.exports = StremioRouter;
```
tymmesyde: Remove the `StremioRouter` line and export `withAuthCheck` instead and rename the file to `withAuthCheck.js`
```js
module.exports = withAuthCheck;
```

src/router/index.js
```diff
@@ -4,10 +4,12 @@ const { useRouteFocused } = require('./RouteFocusedContext');
...
+// const StremioRouter = require('./Router/StremioRouter');
 
 module.exports = {
     useRouteFocused,
     useModalsContainer,
     Modal,
-    Router
+    Router,
+    // StremioRouter
 };
```
tymmesyde: Require `withAuthCheck` and export it
```js
const withAuthCheck = require('./withAuthCheck');

module.exports = {
    // ...
    withAuthCheck,
};

vczb:
This implementation using a HOC on the Router files requires a refactor, there is some errors throwing on the console and the page does not render properly. 

In addition to that, the method `withCoreSuspender` is used on a lot of pages and hooks internaly, wrapping it on the Router requires to remove it from those places

---

I will re-submit my first commit, with fix it in a not scalable but simple way ( tested on the PR description video )

---

Thanks for all who reviewed this PR and support me 

---

For the mantainers, in case of this solution has not be suficient feel free to close this PR and unasign-me of the issue.

---

I will keep following this amazing project, maybe I can help more at the future

tymmesyde:
Thank you for your PR, closed in favor of #429

PR #363, opened 2023-05-17
fix(YoutuberLogo): youtuber logo now shows up

### LINKED ISSUE #362: Youtuber profiles do not show up properly
Banners aren't displayed properly

Mr Beast
image

Dude Perfect
image

Iheuzio:
closes #362 

Youtuber logos will now show up in the info menu. Tried implementing banners but requires no-cores policy, so gave up on that.

image

### DIFF
--- a/src/routes/Player/InfoMenu/InfoMenu.js
+++ b/src/routes/Player/InfoMenu/InfoMenu.js
@@ -29,7 +29,7 @@ const InfoMenu = ({ className, ...props }) => {
                         className={styles['meta-preview']}
                         compact={true}
                         name={metaItem.name}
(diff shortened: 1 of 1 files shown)


nklhtv:
~~Hi, good idea. Let's use poster only in cases where logo is not available.
Also this must be applied in the MetaPreview in Discover.~~
Changed my mind. This should be fixed in the YouTube addon itself. Posters often times have 1x1.5 aspect ratio which will not look good in that place.
Screenshot 2023-05-23 at 23 03 47

TRtomasz:
@Iheuzio  thanks for this PR but we decided to close this as we decided internally to fix it in a different way.

PR #365, opened 2023-05-18
subtitles request wait for video params

unclekingpin:
Depends on https://github.com/Stremio/stremio-video/pull/40


elpiel (approved):

TRtomasz (changes requested):
i tried it out and it makes some weird stuff with subs for some specific videos i have none, like button is greyed out (extraction 2) 
for rest i had only embedded subs and no external subs at all 
i cchecked network and i don't see requests even being made to the addon.

PR #375, opened 2023-05-29
Allow Using External Players in Settings and Player

jaruba:
Untested, awaiting `stremio-core` changes, based on the discussion had today with @unclekingpin and @elpiel

### DIFF
--- a/src/routes/Settings/useProfileSettingsInputs.js
+++ b/src/routes/Settings/useProfileSettingsInputs.js
@@ -158,7 +158,10 @@ const useProfileSettingsInputs = (profile) => {
         }
     }), [profile.settings]);
     const playInExternalPlayerSelect = React.useMemo(() => ({
-        options: externalPlayerOptions,
+        options: externalPlayerOptions.map(opt => {
+            opt.label = t(opt.label);
+            return opt;
+        }),
         selected: [`${profile.settings.playerType || 'internal'}`],
         onSelect: (event) => {
             core.transport.dispatch({
--- a/src/routes/Settings/useProfileSettingsInputs.js
+++ b/src/routes/Settings/useProfileSettingsInputs.js
@@ -213,7 +213,7 @@ const useProfileSettingsInputs = (profile) => {
         }
     }), [profile.settings]);
     const playInExternalPlayerCheckbox = React.useMemo(() => ({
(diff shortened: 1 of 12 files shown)


src/routes/Settings/useProfileSettingsInputs.js
```diff
@@ -213,15 +213,15 @@ const useProfileSettingsInputs = (profile) => {
...
+        checked: profile.settings.playerType === 'external',
         onClick: () => {
             core.transport.dispatch({
                 action: 'Ctx',
                 args: {
                     action: 'UpdateSettings',
                     args: {
                         ...profile.settings,
-                        playInExternalPlayer: !profile.settings.playInExternalPlayer
+                        playerType: profile.settings.playerType !== 'external' ? 'external' : 'internal'
```
sleeyax: I don't have the whole context here, but possible typo?

```suggestion
                        playerType: profile.settings.playerType === 'external' ? 'external' : 'internal'
```

sleeyax: never mind I commented too soon. this is probably supposed to be flipped.

jaruba: it needs to update the value to the opposite of the current value (like a toggle), so the code is already correct, i think

jaruba:
Tested on:

[MacOS] Disabled, VLC (with and without server attached, all is working as expected)
[iPad] Disabled, VLC, Outplayer, Infuse (Infuse is the only one that is acting wonky and refuses to play some streams, but works with others, we may consider disabling support for it)
[Android] Disabled, Allow Choosing, VLC, Just Player, MX Player (all are working as expected)

elpiel (approved):

elpiel:
LGTM!

elpiel:
Stremio-web settings were tested:

- Andoird (firefox) Has options: Allow choosing, VLC Just player MX Player, tested with: Allow choosing, VLC
- Ubuntu 20.04 (firefox), has options: VLC , tested with: VLC

With both cases covered:
- With a running server.js
- Without a running server.js

TRtomasz (approved):
Tested on Android
 and Windows
Works fine, except one thing that has been opened https://github.com/Stremio/stremio-web/issues/378 but not a biggy so can be done after initial release

PR #380, opened 2023-05-31
Open player stream changes

tymmesyde:


TRtomasz:
is this testable? what does it improve?

PR #382, opened 2023-06-01
Fix iPad Detection

jaruba:
iPad was detected as MacOS on Safari


TRtomasz (approved):
still working fine on Android Windows and Mac dont have ipad to test it.

tymmesyde (approved):
Tested on Linux with Chrome / Firefox and Android also with both

PR #385, opened 2023-06-02
Remove Infuse From iOS Ext Players

jaruba:
There are still some bugs with Infuse, it doesn't always start playback.

Should not be available until fixed. (most probably on the Infuse side)


core1024 (approved):

Doctor2023:
IMG_0624

I can't reproduce with infuse I get this message playback

PR #407, opened 2023-06-16
fix: Change copyright to 2023

elpiel:
The test for copyrighting is broken due to non-updated files, see this build for example: https://github.com/Stremio/stremio-web/actions/runs/5224291885/jobs/9432313314


elpiel:
Duplicate of #402

PR #418, opened 2023-06-30
add a long press option to the button module based on timeout. 

dexter21767-dev:
this makes a long press on mobile browser execute a function instead of the default popup menu when long-pressing an anchor. (fully working on ios safari and the PWA)

it's currently only used in video objects (series episodes) for watching or marking as watched. 

documentation regarding the added CSS: [Safari CSS Properties](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariCSSRef/Articles/StandardCSSProperties.html).

it was only tested on:
- ios safari : working as expected.
- ios PWA: working as expected.
- ios firefox: working as expected.
- ios chrome: working as expected, but the context menu of the thumbnail image appears (bad UX).
- windows chrome: working as expected.
- windows chrome with device mode: the "mark as watched" appears, but will disappear if stopped pressing directly without moving the mouse slightly first.

### DIFF
--- a/src/common/Button/Button.js
+++ b/src/common/Button/Button.js
@@ -5,7 +5,19 @@ const PropTypes = require('prop-types');
 const classnames = require('classnames');
 const styles = require('./styles');
 
-const Button = React.forwardRef(({ className, href, disabled, children, ...props }, ref) => {
+const Button = React.forwardRef(({ className, href, disabled, children, onLongPress, ...props }, ref) => {
+    let pressTimer = null;
+    const onTouchStart = function () {
+        pressTimer = setTimeout(function () {
+            if (typeof onLongPress === 'function') {
+                //alert('longpress detected');
+                onLongPress();
+            }
+        }, 600); // values less than 600 will cause an artifact of previous menus staying on screen.
+    };
+    const onTouchEnd = function () {
+        clearTimeout(pressTimer);
+    };
     const onKeyDown = React.useCallback((event) => {
         if (typeof props.onKeyDown === 'function') {
             props.onKeyDown(event);
@@ -36,7 +48,9 @@ const Button = React.forwardRef(({ className, href, disabled, children, ...props
             className: classnames(className, styles['button-container'], { 'disabled': disabled }),
             href,
             onKeyDown,
-            onMouseDown
+            onMouseDown,
+            onTouchStart,
+            onTouchEnd,
         },
         children
     );
@@ -50,7 +64,8 @@ Button.propTypes = {
     disabled: PropTypes.bool,
     children: PropTypes.node,
     onKeyDown: PropTypes.func,
-    onMouseDown: PropTypes.func
+    onMouseDown: PropTypes.func,
+    onLongPress: PropTypes.func,
 };
 
 module.exports = Button;
--- a/src/common/Button/Button.js
+++ b/src/common/Button/Button.js
@@ -8,6 +8,9 @@ const { useLongPress } = require('use-long-press');
 
 const Button = React.forwardRef(({ className, href, disabled, children, onLongPress, ...props }, ref) => {
     const longPress = useLongPress(onLongPress);
+    // exoposing them to make it easier to know which handlers are exported,
+    // in case a change to one of them is needed in the future.
+    const {onPointerDown, onPointerMove, onPointerUp, onPointerLeave} = longPress();
     const onKeyDown = React.useCallback((event) => {
         if (typeof props.onKeyDown === 'function') {
             props.onKeyDown(event);
@@ -39,7 +42,7 @@ const Button = React.forwardRef(({ className, href, disabled, children, onLongPr
             href,
             onKeyDown,
             onMouseDown,
-            ...longPress()
+            onPointerDown, onPointerMove, onPointerUp, onPointerLeave
         },
         children
     );
--- a/src/common/Button/Button.js
+++ b/src/common/Button/Button.js
@@ -13,7 +13,7 @@ const Button = React.forwardRef(({ className, href, disabled, children, onLongPr
                 //alert('longpress detected');
                 onLongPress();
             }
-        }, 600); // values less than 600 will cause an artifact of previous menus staying on screen.
+        }, 650); // an artifact of previous menus staying on the screen will happen on Safari if the timeout was set to 600 and less, and 650 for PWA.
     };
     const onTouchEnd = function () {
         clearTimeout(pressTimer);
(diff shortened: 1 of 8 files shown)


src/common/Button/Button.js
```diff
@@ -5,7 +5,19 @@ const PropTypes = require('prop-types');
 const classnames = require('classnames');
 const styles = require('./styles');
 
-const Button = React.forwardRef(({ className, href, disabled, children, ...props }, ref) => {
+const Button = React.forwardRef(({ className, href, disabled, children, onLongPress, ...props }, ref) => {
+    let pressTimer = null;
```
tymmesyde: Should probably use a Ref here:
```js
const longPressTimeout = React.useRef(null);
```

src/common/Button/Button.js
```diff
@@ -5,7 +5,19 @@ const PropTypes = require('prop-types');
...
+    const onTouchStart = function () {
+        pressTimer = setTimeout(function () {
```
tymmesyde: Maybe add a clearTimeout too here to cancel the previous one
```js
clearTimetout(longPressTimeout.current);
longPressTimeout.current = setTimeout(...
```

src/common/Button/Button.js
```diff
@@ -5,7 +5,19 @@ const PropTypes = require('prop-types');
 const classnames = require('classnames');
 const styles = require('./styles');
 
-const Button = React.forwardRef(({ className, href, disabled, children, ...props }, ref) => {
+const Button = React.forwardRef(({ className, href, disabled, children, onLongPress, ...props }, ref) => {
+    let pressTimer = null;
+    const onTouchStart = function () {
```
tymmesyde: This function should probably be a useCallback with `onLongPress` as dependency

tymmesyde (changes requested):

tymmesyde:
Tested on Android / Desktop with Firefox and Chrome

tymmesyde (dismissed):

src/common/Button/Button.js
```diff
@@ -36,7 +47,9 @@ const Button = React.forwardRef(({ className, href, disabled, children, ...props
             className: classnames(className, styles['button-container'], { 'disabled': disabled }),
             href,
             onKeyDown,
-            onMouseDown
+            onMouseDown,
+            onTouchStart,
+            onTouchEnd,
```
unclekingpin: Should we also handle touchcancel?

src/common/Button/Button.js
```diff
@@ -5,7 +5,18 @@ const PropTypes = require('prop-types');
 const classnames = require('classnames');
 const styles = require('./styles');
 
-const Button = React.forwardRef(({ className, href, disabled, children, ...props }, ref) => {
+const Button = React.forwardRef(({ className, href, disabled, children, onLongPress, ...props }, ref) => {
+    const longPressTimeout = React.useRef(null);
+    const onTouchStart = React.useCallback((event) => {
+        longPressTimeout.current = setTimeout(function () {
```
unclekingpin: must clear the timeout before starting a new one

src/routes/MetaDetails/VideosList/Video/Video.js
```diff
@@ -29,6 +30,15 @@ const Video = ({ className, id, title, thumbnail, episode, released, upcoming, w
             toggleMenu();
         }
     }, [toggleMenu]);
+    const popupLabelOnLongPress = React.useCallback((event) => {
+        if (!event.nativeEvent.togglePopupPrevented && !event.nativeEvent.ctrlKey) {
+            if (event.cancelable) event.preventDefault();
+            toggleMenu();
+        }
+    }, [toggleMenu]);
+    const popupMenuOnLongPress = React.useCallback((event) => {
```
unclekingpin: this must be named with 'onTouchStart' suffix instead

src/common/Button/styles.less
```diff
@@ -3,11 +3,19 @@
 @import (reference) '~@stremio/stremio-colors/less/stremio-colors.less';
 
 .button-container {
+    // IOS specific 
+    // prevents showing the default context-menu when long pressing an anchor in safari.
+    -webkit-touch-callout: none !important;
+    // prevents user from selecting text from a div on long pressing in safari.
+    -webkit-user-select: none !important;
+
```
unclekingpin:
```suggestion
```

src/common/Button/styles.less
```diff
@@ -3,11 +3,19 @@
 @import (reference) '~@stremio/stremio-colors/less/stremio-colors.less';
 
 .button-container {
+    // IOS specific 
+    // prevents showing the default context-menu when long pressing an anchor in safari.
+    -webkit-touch-callout: none !important;
+    // prevents user from selecting text from a div on long pressing in safari.
+    -webkit-user-select: none !important;
```
unclekingpin: is that not handled in app's [styles](https://github.com/Stremio/stremio-web/blob/development/src/App/styles.less#L36)?
if not, is it a better place for it?
does the build system handle browser prefixing for us?

dexter21767-dev: `-webkit-user-select: none;` can be moved to the app's styles.
but I don't know if moving `-webkit-touch-callout` there is a good idea, as this will disable the default context menu, and I dunno if it's needed (for example in the addons section). if it's not needed anywhere I can move it to the app's styles.

src/routes/MetaDetails/VideosList/Video/styles.less
```diff
@@ -12,6 +12,7 @@
 }
 
 .video-container {
+
```
dexter21767-dev: as for the spaces and empty lines, I left them as they were and didn't change them. I think changing them in a different PR would be better, but I will change them if you think I should.

src/common/Button/Button.js
dexter21767-dev: i don't think it's necessary. even after multiple testing I couldn't make it trigger the touchcancel event.
anything shorter than a 'longpress' will default to a normal press and open the streams menu.

src/routes/MetaDetails/VideosList/Video/styles.less
unclekingpin:
```suggestion
```

src/common/Button/styles.less
```diff
@@ -7,7 +7,7 @@
     outline-color: @color-surface-light5;
     outline-offset: calc(-1 * var(--focus-outline-size));
     cursor: pointer;
-
+
```
unclekingpin:
```suggestion
```

src/common/Button/styles.less
```diff
@@ -7,7 +7,6 @@
     outline-color: @color-surface-light5;
     outline-offset: calc(-1 * var(--focus-outline-size));
     cursor: pointer;
-
     &:focus {
```
unclekingpin:
```suggestion

    &:focus {
```

unclekingpin (approved):

PR #419, opened 2023-06-30
fix: add missing prop to the discover preview

0xRnato:
There was a missing prop needed to display imdb rating, genres, cast and directors on the discover preview.

image


TRtomasz:
seems fine

TRtomasz (approved):

jaruba:
@0xRnato thanks!

PR #420, opened 2023-07-03
feat: Add ability to start with custom streaming server

### LINKED ISSUE #398: Set Stremio Web Server With Querystring
When accessing the server via http, it redirects to the old UI ( https://app.strem.io/shell-v4.4/?streamingServer=streamingserverhere)
How do I make it redirect to web.strem.io with the server being hardcoded on the url, overwriting the user's server in the settings?

Thanks.

felipemfp:
This PR allows us to overwrite the default streaming server using the `streamingServerUrl` in the URL. For example:

> http://localhost:8080/?streamingServerUrl=http%3A%2F%2F192.168.1.12%3A11470


TRtomasz:
This works partially only
1. It works only on init/refresh if i simply change port in the url and press enter, server address will not be changed in the app i have to press f5
2. There is missing notification like in current Stremio 4 version we have `useToast` https://github.com/Stremio/stremio-web/tree/d44f7047dcee5f0b2257318ac7153a09bbfe2959/src/common/Toast here this can be used

TRtomasz (changes requested):
explained in comment above.

src/App/DefaultSettingsHandler.js
```diff
@@ -0,0 +1,25 @@
+// Copyright (C) 2017-2023 Smart code 203358507
+
+const React = require('react');
+const {
+    withCoreSuspender,
+    useProfile,
+} = require('stremio/common');
+const useProfileSettingsInputs = require('stremio/routes/Settings/useProfileSettingsInputs');
```
tymmesyde: I would suggest to use the core service transport to update settings instead of importing useProfileSettingsInputs and add `profile.settings` as a dependency to the hook
```js
const { useServices } = require('stremio/services');

const DefaultSettingsHandler = () => {
    const profile = useProfile();
    const { core } = useServices();
  
    React.useEffect(() => {
        const searchParams = new URLSearchParams(window.location.search);
        if (searchParams.has('streamingServerUrl')) {
           core.transport.dispatch({
                action: 'Ctx',
                args: {
                    action: 'UpdateSettings',
                    args: {
                        ...profile.settings,
                        streamingServerUrl: searchParams.get('streamingServerUrl')
                    }
                }
            });
        }
    }, [profile.settings]);
 
    return null;
};
```

src/App/DefaultSettingsHandler.js
```diff
@@ -0,0 +1,51 @@
...
+                    action: 'StreamingServer',
+                    args: {
+                        action: 'Reload'
+                    }
+                });
+            }, 1000);
+            toast.show({
+                type: 'info',
+                title: `Using streaming server at ${streamingServerUrl}`,
+                timeout: 4000
```
gaurovgiri, line 43:
```suggestion
            core.transport.dispatch(updateSettingsAction).then(() => {
                // Dispatch the 'StreamingServer' action after the 'UpdateSettings' action is completed.
                const streamingServerAction = {
                    action: 'StreamingServer',
                    args: {
                        action: 'Reload',
                    },
                };
                return core.transport.dispatch(streamingServerAction);
            })
            .then(() => {
                toast.show({
                    type: 'info',
                    title: `Using streaming server at ${streamingServerUrl}`,
                    timeout: 4000,
                });
            })
            .catch((error) => {
                // Handle any errors here.
                console.error('Error:', error);
```

gaurovgiri (changes requested):
Wrap the second dispatch ('StreamingServer' action) in a .then() block of the first dispatch to ensure that it happens only after the first action is completed.

Add a .catch() block to handle any potential errors that might occur during the dispatching of actions.

mpetuska:
Could we get this merged please? This is the last missing piece from my homeserver setup.

kKaskak:
> Could we get this merged please? This is the last missing piece from my homeserver setup.

hi this will be merged and included in the next stremio-web release

PR #424, opened 2023-07-11
Mark Video as Watched When Using External Player

jaruba:


TRtomasz (dismissed):
seems to be working although not sure if i'm a fan of marking as watched right after you start playing something

jaruba:
@TRtomasz there is no other option except marking it as watched immediately, any other case is error prone, maybe the user won't even go back to the web app and will close the entire browser, then it won't be marked as watched at all

tymmesyde:
Tested and works, maybe we should also do that for the player feature ?
https://github.com/Stremio/stremio-web/blob/development/src/routes/Player/OptionsMenu/OptionsMenu.js#L56

And maybe it's better in that case to do that when we receive the event `PlayingOnDevice` from core instead of putting functions in each components: 
https://github.com/Stremio/stremio-web/blob/development/src/App/App.js#L97

unclekingpin:
> Tested and works, maybe we should also do that for the player feature ? https://github.com/Stremio/stremio-web/blob/development/src/routes/Player/OptionsMenu/OptionsMenu.js#L56
> 
> And maybe it's better in that case to do that when we receive the event `PlayingOnDevice` from core instead of putting functions in each components: https://github.com/Stremio/stremio-web/blob/development/src/App/App.js#L97


That could be implemented within core itself

TRtomasz:
@unclekingpin  does it require retesting? i see no logic has changed after your commits

jaruba:
@TRtomasz we should always re-test for safety

TRtomasz (changes requested):
hm, it works when opening in vlc but i no longer see option to mark it as watched/unwatched manually, i'm like 50% sure it was there before

tymmesyde:
> hm, it works when opening in vlc but i no longer see option to mark it as watched/unwatched manually, i'm like 50% sure it was there before

This bug was introduced in a previous PR and only happens on Firefox, see #428

jaruba:
@TRtomasz you mean this? https://github.com/Stremio/stremio-web/pull/418

TRtomasz:
@jaruba  
image
 this, it is gone with this PR, couldn't open this menu
...
checked once again, yes context menu with this function is gone now

TRtomasz (dismissed):
approving as context menu is fixed here https://github.com/Stremio/stremio-web/pull/428

tymmesyde:
> > Tested and works, maybe we should also do that for the player feature ? https://github.com/Stremio/stremio-web/blob/development/src/routes/Player/OptionsMenu/OptionsMenu.js#L56
> > And maybe it's better in that case to do that when we receive the event `PlayingOnDevice` from core instead of putting functions in each components: https://github.com/Stremio/stremio-web/blob/development/src/App/App.js#L97
> 
> That could be implemented within core itself

The only way i can see this implemented easily in core is by passing the video id to the PlayOnDevice action as well, not sure if this is a good design since it has nothing to do with the streaming server

elpiel (approved):

PR #428, opened 2023-07-18
fix(Video): context menu was not working on Firefox

tymmesyde:
~~The context menu was not working on Firefox because the `contextmenu` event is not a PointerEvent on Firefox and we were using the pointerEvent property to check if it was from a mouse
https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event#browser_compatibility~~

~~The `contextmenu` event is managed by the browser, on desktop it should trigger on right click, on mobile on long press
So we don't really need to use the  `onLongPress` callback~~

We ended up using `mouseup` instead as it's a MouseEvent and check if right click was pressed
`onMouseUp` for desktop and `onLongPress` for mobile

### DIFF
--- a/src/routes/MetaDetails/VideosList/Video/Video.js
+++ b/src/routes/MetaDetails/VideosList/Video/Video.js
@@ -24,13 +24,6 @@ const Video = ({ className, id, title, thumbnail, episode, released, upcoming, w
     const popupLabelOnContextMenu = React.useCallback((event) => {
         if (!event.nativeEvent.togglePopupPrevented && !event.nativeEvent.ctrlKey) {
             event.preventDefault();
-            if (event.nativeEvent.pointerType === 'mouse') {
-                toggleMenu();
-            }
-        }
-    }, [toggleMenu]);
-    const popupLabelOnLongPress = React.useCallback((event) => {
-        if (event.nativeEvent.pointerType !== 'mouse' && !event.nativeEvent.togglePopupPrevented) {
             toggleMenu();
         }
     }, [toggleMenu]);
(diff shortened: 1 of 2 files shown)


unclekingpin:
There must be a better way to solve this because that will break UX on mobile safari.
Can we check if the nativeEvent is instance of PointerEvent and if its not then threat it as a mouse event
cc: @dexter21767-dev

unclekingpin:
```js
     const popupLabelOnContextMenu = React.useCallback((event) => {
        if (!event.nativeEvent.togglePopupPrevented && !event.nativeEvent.ctrlKey) {
            event.preventDefault();
            if (typeof event.nativeEvent.pointerType !== 'string' || event.nativeEvent.pointerType === 'mouse') {
                toggleMenu();
            }
        }
    }, [toggleMenu]);
    const popupLabelOnLongPress = React.useCallback((event) => {
        if (!event.nativeEvent.togglePopupPrevented && typeof event.nativeEvent.pointerType === 'string' && event.nativeEvent.pointerType !== 'mouse') {
            toggleMenu();
        }
    }, [toggleMenu]);
```

TRtomasz (changes requested):
this indeed fix context menu but i noticed this in the console
image
Also for some reason i have no streams for movies using this build, not sure if it got broken with some other change

tymmesyde:
> this indeed fix context menu but i noticed this in the console image Also for some reason i have no streams for movies using this build, not sure if it got broken with some other change

The error was introduced in this PR: https://github.com/Stremio/stremio-web/pull/375
It's a prop check error so nothing too alarming

unclekingpin (dismissed):

TRtomasz (approved):

PR #438, opened 2023-08-02
feat: add notifications toggle on details

tymmesyde:
Need https://github.com/Stremio/stremio-core/pull/486 and https://github.com/Stremio/stremio-core-web/pull/58
Waiting for core-web to be updated
Related to https://github.com/Stremio/stremio-web/pull/437

### DIFF
--- a/src/types/models/Ctx.d.ts
+++ b/src/types/models/Ctx.d.ts
@@ -45,8 +45,14 @@ type Profile = {
 type Notifications = {
     uid: string,
     created: string,
-    items: Record<string, Video[]>,
+    items: Record<string, NotificationItem[]>,
 };
+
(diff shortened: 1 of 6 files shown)


src/types/models/Ctx.d.ts
```diff
@@ -42,6 +42,13 @@ type Profile = {
     settings: Settings,
 };
 
+type Notifications = {
+    uid: string,
+    created: string,
+    items: Record<string, Video[]>,
```
elpiel: This is no more a Video. Instead it's a NotificationItem with videoId, metaId & videoReleased.
- [x] Update (See #440 )

jaruba (approved):

PR #439, opened 2023-08-10
fix: Board has now it's own items instead of LibraryItem

elpiel:
Requires:
- [ ] core - https://github.com/Stremio/stremio-core/pull/494
- [ ] core-web - https://github.com/Stremio/stremio-core-web/pull/60


tymmesyde:
Merged in #437 and  #438

PR #461, opened 2023-09-05
Description bug fix and test file deleted

kKaskak:
Description not showing up in the movie/serie details page not showing up fixed

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+MetaPreview.js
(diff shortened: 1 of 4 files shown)


tymmesyde (dismissed):

tymmesyde (commented):
Now that i think about it, we could make it look like the desktop version: under the tags with the summary title on the Detail page
But keep it like this for the compact version like on the Discover page

tymmesyde (approved):

PR #472, opened 2023-10-10
Actually show streaming server error

edde746:
title


tymmesyde:
Thank you !
Should probably just be 
```
t('SETTINGS_SERVER_STATUS_ERROR')
```
as the error message is more dev oriented and in english

tymmesyde (approved):

PR #482, opened 2023-10-17
add rewind buttons

thedanielius:
Added two rewind buttons(Netflix like).

+10s and -10s.

Definitely needs new icons added [here](https://stremio.github.io/stremio-icons/).

Would be a neat feature for Android app also.

image


unclekingpin (commented):
We have skip-15-forward and skip-15-back icon. try those icons instead.

PR #483, opened 2023-10-18
add infuse player for ios

### LINKED ISSUE #463: Allow Infuse as an external player on iOS and iPadOS
No option to select Infuse as an external player.

lucasmarcondes:
Adds infuse to the list of external players for iOS. 

PR will resolve #463

### DIFF
--- a/src/common/externalPlayerOptions.js
+++ b/src/common/externalPlayerOptions.js
@@ -7,7 +7,8 @@ let options = [{ label: 'EXTERNAL_PLAYER_DISABLED', value: 'internal' }];
 if (platform.name === 'ios') {
     options = options.concat([
         { label: 'VLC', value: 'vlc' },
-        { label: 'Outplayer', value: 'outplayer' }
+        { label: 'Outplayer', value: 'outplayer' },
+        { label: 'Infuse', value: 'infuse' }
     ]);
 } else if (platform.name === 'android') {
     options = options.concat([


tymmesyde:
We are not able to support infuse player, but thank you for your PR

prom3theu5:
I think infuse really needs to be looked at again

I changed `common/platform.js` to always return ios in the version of 5.0.0.beta.0 I run on my server, and added infuse back to `common/externalPlayerOptions.js` for iOS,as support for it was never removed from core - its only been removed from the web external players menu.

Just finished watching 152 episodes of a certain series over the past 8 days with it auto launching infuse as the external player, and haven't experienced the playback error once.

About 20 episodes were watched on my iphone, about 20 on my iPad, and the rest on my mac running Sonoma.
I run infuse (with pro subscription) on all devices

So - i'm fairly certain whatever error they had - has now been fixed, unless the error is related to transcoding via server.js?
As non of my media files were transcoded by stremio server

prom3theu5:
Scratch that - I think I know why you dont support it
Infuse doesn't appear to support m3u files

The way I've been watching things (with a custom addon im playing with) pulls from jellyfin, and gives direct media url - not an m3u playlist

lucasmarcondes:
@tymmesyde, I read somewhere on Reddit that there is a bug within Infuse, which is why Stremio doesn’t fully support it. 

Infuse released an update to their docs recently and their apps that allows for “third party apps to play, bookmark, and save videos”. Is this related to the fix Stremio needs? 

https://support.firecore.com/hc/en-us/articles/215090997-API-for-Third-Party-Apps-Services

PR #487, opened 2023-10-18
fix-playback-slider-circle

kKaskak:


tymmesyde (changes requested):
The slider component is also used for the volume control
So the purple circle is also present here

tymmesyde (approved):

PR #491, opened 2023-10-20
Version label was almost not visible

tymmesyde:


src/routes/Settings/styles.less
```diff
@@ -69,7 +69,8 @@
             .version-info-label {
                 flex: 0 1 auto;
                 margin: 0.5rem 0;
-                color: var(--overlay-color);
+                color: var(--primary-foreground-color);
+                opacity: 0.5;
```
gaurovgiri:
```suggestion
                opacity: 0.3;
```

gaurovgiri (changes requested):
Setting opacity to 0.3 will help distinguish the version info from the clickable buttons in settings which are of opacity 0.4.

kKaskak (changes requested):
It looks good with opacity 0.3, but it might require display changes on mobile. As it's brighter than before and fixed with the Navbar. What if we just put it in the same place as before but not fixed, visible only when you scroll down the settings and add some margin after it, so there is enough spacing between the label and the Navbar.

gaurovgiri:
@kKaskak Yeah, I think doing so might be better.

src/routes/Settings/styles.less
```diff
@@ -465,6 +465,10 @@
                         }
                     }
                 }
+
```
kKaskak, line 468: @tymmesyde You can remove this as it works fine without specifying flex in the minimum size too 👍

src/routes/Settings/styles.less
```diff
@@ -438,18 +449,7 @@
             flex-direction: column-reverse;
 
             .side-menu-container {
```
kKaskak, line 451: remove this too as its already specified with display: none on a higher resolution size

kKaskak (changes requested):

src/routes/Settings/styles.less
kKaskak: 468-471

src/routes/Settings/styles.less
kKaskak: 451-453

tymmesyde: It's better to make it clear what do we want to change for each screen sizes

kKaskak (approved):

PR #493, opened 2023-10-22
translating type in board catalog titles

### LINKED ISSUE #453: Use translations strings on board for content type
**Description**
On board we do not use translations strings for content type. In stremio 4 we translate names like Movie or Series using translations but in stremio 5 they are always displayed in english.

VictorLutes:
Uses translation for the catalog type (ie: Series, Movie, Channel, ...) in Board titles

PR will resolve https://github.com/Stremio/stremio-web/issues/453

### DIFF
--- a/src/routes/Board/Board.js
+++ b/src/routes/Board/Board.js
@@ -4,6 +4,7 @@ const React = require('react');
 const classnames = require('classnames');
 const debounce = require('lodash.debounce');
 const { useTranslation } = require('react-i18next');
+const { translateOption } = require('stremio/common');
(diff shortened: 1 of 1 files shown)


gaurovgiri (approved):
LGTM

unclekingpin (changes requested):
the issue is in stremio-core-web and the fix should start [from there](https://github.com/Stremio/stremio-core-web/blob/development/src/model/serialize_catalogs_with_extra.rs#L67C21-L75C23)
the `.title` should be replaced by `.type` and `.name` and the catalog title should be build in js

tymmesyde:
Closed in favor of https://github.com/Stremio/stremio-web/pull/540
Thank you for your contribution

PR #513, opened 2023-11-22
Add IINA & mpv as external players

edde746:
Depends on https://github.com/Stremio/stremio-core/pull/564


edde746:
To explain the second commit, it cleans up the code and makes use of the users selected external player even if they are on desktop and have a working streaming server (previously it would disregard the selected external player unless the user was either on mobile or did not have a streaming server).

tymmesyde:
~~To support another external player, you need to add the uri schemes to the core
https://github.com/Stremio/stremio-core/blob/development/src/deep_links/mod.rs#L77~~

Missed the link about your core PR

src/common/externalPlayerOptions.js
```diff
@@ -16,7 +16,12 @@ if (platform.name === 'ios') {
         { label: 'Just Player', value: 'justplayer' },
         { label: 'MX Player', value: 'mxplayer' }
     ]);
-} else if (['windows', 'macos', 'linux'].includes(platform.name)) {
+} else if (platform.name === 'macos') {
+    options = options.concat([
+        { label: 'IINA', value: 'iina' },
+        { label: 'mpv', value: 'mpv' }
+    ]);
```
elpiel: Until now VLC would have been an option for Macos but since your change only IINA & mpv are, so we need to add VLC back to the list.

- [ ] Please add the VLC player here as well.

elpiel (changes requested):
Thank you very much for your contribution!
This is a great addition to stremio-web.

src/common/externalPlayerOptions.js
edde746: I'm pretty sure VLC does not work, but sure.

kKaskak (changes requested):
While we are already on this, can we also add translation to the default option?

edde746:
@kKaskak Are you sure you are commenting on the right PR? I have no idea what you mean since nothing in this PR is language specific/can be translated...

kKaskak:
> @kKaskak Are you sure you are commenting on the right PR? I have no idea what you mean since nothing in this PR is language specific/can be translated...
https://github.com/Stremio/stremio-web/blob/39bdfba41d94ca498431e5fa99080a5dee46304a/src/common/externalPlayerOptions.js#L5

This should most likely be translated.

edde746:
> This should most likely be translated.

[It already is?](https://github.com/Stremio/stremio-web/blob/development/src/routes/Settings/useProfileSettingsInputs.js#L200)

kKaskak:
> > This should most likely be translated.
> 
> [It already is?](https://github.com/Stremio/stremio-web/blob/development/src/routes/Settings/useProfileSettingsInputs.js#L200)

I tested it and it didn't translate that's why. 🤔 Will test again. If it works correctly -> approving.

kKaskak:
> > This should most likely be translated.
> 
> [It already is?](https://github.com/Stremio/stremio-web/blob/development/src/routes/Settings/useProfileSettingsInputs.js#L200)

In fact the translation was not implemented correctly, as you can see in the screenshots below.  It was translating the options only after the whole page was refreshed. Pushed a fix to the branch.
Screenshot 2023-11-27 at 11 40 00
Screenshot 2023-11-27 at 11 39 43

kKaskak (dismissed):

elpiel (dismissed):

tymmesyde (approved):

PR #514, opened 2023-11-22
Implement remote https endpoint settings

tymmesyde:
Close https://github.com/Stremio/stremio-tasks/issues/97
Need https://github.com/Stremio/stremio-server/pull/201


kKaskak (approved):

TRtomasz:
Approving since setting is added but missing things are:
1. Default setting is empty not disabled
2. It redirect to stremio 4 instead stremio web

TRtomasz (approved):

PR #521, opened 2023-11-29
chore: streamState change for subtitles and audio tracks

elpiel:
We send partial state updates which is not handled by core.
Core only applies changes to the full state.


tymmesyde:
Closed the streams-state branch

PR #526, opened 2023-12-04
Improve docker image

edde746:
- Removes unnecessary `http_server.js`
- Reduces size from ~970 MB to 18 MB

Bechmark with 100k requests:

Before
```
Summary:
  Total:	14.0344 secs
  Slowest:	0.0427 secs
  Fastest:	0.0006 secs
  Average:	0.0070 secs
  Requests/sec:	7125.3674
```

After
```
Summary:
  Total:	4.8001 secs
  Slowest:	0.0404 secs
  Fastest:	0.0002 secs
  Average:	0.0024 secs
  Requests/sec:	20832.8811
```


unclekingpin (approved):
Nice! The only difference is that we dont have a custom 404 page.

unclekingpin (changes requested):
actually we need the 404 handler for sure, which we will customize later

edde746:
The current dockerfile does not have a custom 404 page either, it just has a comment saying it should be added later. To add a custom 404 page, a `404.html` would need to be created and passed as the [fallback argument to goStatic](https://github.com/PierreZ/goStatic#usage).

prom3theu5:
Have you considered using caddy instead?
That way you can make it super super simple for people to also auto gen tls etc?
https://caddyserver.com/

Also has a 17mb base image size for caddy
https://hub.docker.com/_/caddy

edde746:
TLS should be handled by the user IMO, gives them control of how they want to do it. The purpose of the Docker image should just be to serve the static files and nothing more.

As for the size, while ~16 MB difference does not matter, the base size of `pierrezemb/gostatic` is only ~2 MB. 18 MB is for goStatic + the static files.

prom3theu5:
While I agree with the statement TLS should be handled by the user, I have to add it still would be. They'd still have to mount a caddy file to enable it, configure their load balancing, cache settings, rewrites etc if they wanted to opt into the features. The point is they'd have the option to. 

Just gives users who are hosting in something like azure container apps / instances, where they probably wouldn't want to pay for azure gateway or introduce a second container instance as a load balancer the ability to have easily opt-in features, with the use of a single file server rather than a static file server reverse proxied by a second file server.

That's ok I did say "considered" 😛

kKaskak:
closed due inactivity

PR #530, opened 2023-12-06
Show fullscreen button on Safari desktop PWA

edde746:
With [Safari adding PWA support on desktop](https://developer.apple.com/documentation/safari-release-notes/safari-17-release-notes#Web-apps), `window.navigator.standalone` is no longer only true on iOS.

### DIFF
--- a/src/common/NavBar/HorizontalNavBar/HorizontalNavBar.js
+++ b/src/common/NavBar/HorizontalNavBar/HorizontalNavBar.js
@@ -18,7 +18,6 @@ const HorizontalNavBar = React.memo(({ className, route, query, title, backButto
         window.history.back();
     }, []);
     const [fullscreen, requestFullscreen, exitFullscreen] = useFullscreen();
-    const [isIOSPWA] = usePWA();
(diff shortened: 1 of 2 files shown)


kKaskak:
Screenshot 2023-12-07 at 0 52 14
I personally don't think this change is needed at all, as we already have a full screen button in the native app action buttons, while running PWA on macOS. Leaving the review to @tymmesyde

edde746:
The fullscreen button is completely unnecessary in Stremio web, but this makes the Safari PWA experience consistent with other browsers.

unclekingpin (changes requested):
Can you check for [isMobile](https://github.com/Stremio/stremio-web/blob/development/src/common/platform.js#L25) to hide the fullscreen button

edde746:
On Chrome on Android, you can enter fullscreen mode and on iOS it would be hidden since `document.fullscreenEnabled` is not true. It actually seems that Chrome on Android is one of the only places where the button is useful since there is no F11 key or other way to enter fullscreen mode.

kKaskak:
closed due inactivity

PR #542, opened 2023-12-20
fix: external player stream logic

tymmesyde:
Need https://github.com/Stremio/stremio-core/pull/598


src/common/CONSTANTS.js
```diff
@@ -40,6 +40,49 @@ const ICON_FOR_TYPE = new Map([
...
+    },
+    {
+        label: 'EXTERNAL_PLAYER_ALLOW_CHOOSING',
+        value: 'choose',
+        platforms: ['android'],
+    },
+    {
+        label: 'VLC',
+        value: 'vlc',
+        platforms: ['ios', 'android'],
```
elpiel, line 57:
```suggestion
        platforms: ['ios', 'android', 'macos', 'windows', 'linux'],
```

tymmesyde: We only support VLC for ios and android for this feature

elpiel: As noted by tim, we indeed do not support them.
Also cross-checked with core:
https://github.com/Stremio/stremio-core/blob/bcd037874e1e73990a9688b06adbd551bf59709b/src/deep_links/mod.rs#L85-L91

jaruba (dismissed):

PR #550, opened 2023-12-25
fix: reduced font-size on smaller devices

kKaskak:
## BEFORE
IMG_4099
## NOW
IMG_4100


kKaskak:
this change is mainly because of overflow and to match the current font-size of the stream description

tymmesyde (approved):

PR #555, opened 2024-01-09
fix(MetaDetails): set first season as default

tymmesyde:
Closes https://github.com/Stremio/stremio-tasks/issues/146


TRtomasz (changes requested):
now everything opens on first season. It should open on last watched episode for the stuff you already started watching and first season for not watched series

PR #570, opened 2024-01-29
fix: duplicate local-search items fix

kKaskak:


elpiel (changes requested):
I guess this is not needed anymore since we are de-duplicating in core-web:
https://github.com/Stremio/stremio-core-web/pull/89

PR #599, opened 2024-03-05
feat: add m3u as external player option

tymmesyde:


kKaskak (approved):

TRtomasz:
Should it work without Streaming server? I tried without SS and option was avilalbe in the settings, but when enabling it nothing happens other than error 'video not supported'(Tried on firefox)

TRtomasz (approved):

PR #601, opened 2024-03-06
fix: metaitem titles overflow

kKaskak:
Fixed the issue where the second text line would be partially invisible and cut out, used line-clamp to limit the lines to max 2

### DIFF
--- a/src/common/MetaItem/styles.less
+++ b/src/common/MetaItem/styles.less
@@ -317,11 +317,15 @@
 
         .title-label {
             flex: 1;
-            max-height: 2.4em;
+            max-height: 2.6em;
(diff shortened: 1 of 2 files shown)


src/common/MetaItem/styles.less
```diff
@@ -317,11 +317,15 @@
 
         .title-label {
             flex: 1;
-            max-height: 2.4em;
+            max-height: 2.6em;
```
tymmesyde: `max-height` can be removed

tymmesyde:
```suggestion
```

tymmesyde (approved):

Issue #539, opened 2023-12-19
Bug: Zoom level affects infinite scroll ( can't infiniscroll if you don't have a scroll bar )


Acecool:
If you have a plugin which populates top choices in the web-version of StremIO, and you have a large monitor such as a 4k, and your web-zoom is set to 100% it is possible for the top results to NOT produce a scrollbar. Without a scrollbar, you can't scroll to the bottom, and without being able to scroll to the bottom the infiniscroll system doesn't activate meaning you can't see more results.

If you change your zoom so you end up with a scroll bar, then it works and you can reset the zoom back to your normal level.

kKaskak:
this is fixed

PR #607, opened 2024-03-11
fix: allow scrollToBottom hook to trigger when there is no scroll bar

CrudeCreations:
Hi there, just submitting this PR as I believe it should solve the issue [referenced here](https://github.com/Stremio/stremio-web/issues/539).

Unfortunately I don't have a large enough screen to be able to recreate this issue.

Updating the hook to also check for the target's height to be greater or equal to scroll height should also trigger cb.

I don't know whether you'd want this to be an optional param or whether you think it needs a bit more work.
This may cause more calls to this callback than you'd want if 'hasNextPage' is false and the screen remains unpopulated.


tymmesyde:
Hi ! This will not work because if the `clientHeight` is the same height as `scrollHeight`, the scroll event will not be triggered

kKaskak:
closed due inactivity

PR #612, opened 2024-03-21
fix: Player should use nextStream deep link

elpiel:


tymmesyde:
Closed in favor of changes in core

PR #632, opened 2024-04-22
feature: improve error msg design on metadetails

kKaskak:
## BEFORE 

Screenshot 2024-04-22 at 12 20 05
Screenshot 2024-04-22 at 12 20 18

## AFTER

Screenshot 2024-04-22 at 12 19 38
Screenshot from 2024-05-13 16-40-16

### DIFF
--- a/src/routes/MetaDetails/VideosList/styles.less
+++ b/src/routes/MetaDetails/VideosList/styles.less
@@ -74,9 +74,5 @@
 @media only screen and (max-width: @minimum) {
     .videos-list-container {
         overflow: visible;
-
-        .message-container {
-            display: none;
-        }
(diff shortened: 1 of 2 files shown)


src/routes/MetaDetails/VideosList/styles.less
```diff
@@ -75,5 +74,9 @@
 @media only screen and (max-width: @minimum) {
     .videos-list-container {
         overflow: visible;
+
+        .message-container {
+            display: none;
```
tymmesyde: This probably should not be hidden in case of metadata found but no videos

tymmesyde (approved):

PR #652, opened 2024-06-23
Proxy streams enabled

unclekingpin:
can be tested at https://stremio.github.io/stremio-web/proxy_streams_enabled/
if server settings proxy_streams_enabled is true, then it will proxy all url streams through the /proxy endpoint of that server, regardless of the stream behavior hints.
if server settings proxy_streams_enabled is false, then the stream will be proxied only if there is a behavior hint `proxyHeaders`


unclekingpin:
Its missing translation string for the settings page
изображение

kKaskak:
outdated

PR #658, opened 2024-07-10
feat: copy stream link from streams list

Namyts:
Added ability to copy an Infohash when right-clicking search results.
Currently the way to do this is to wait for the stream to load, pray you can pause the stream before it starts buffering again, and clicking the corresponding button. This can be quite frustrating for slow streams where downloading a stream for external use could be preferable. (Why not use an external search engine? Stremio provides a nice all-in-one-with-addons search)

### DIFF
--- a/src/routes/MetaDetails/StreamsList/Stream/Stream.js
+++ b/src/routes/MetaDetails/StreamsList/Stream/Stream.js
@@ -4,33 +4,14 @@ const React = require('react');
 const PropTypes = require('prop-types');
 const classnames = require('classnames');
 const { default: Icon } = require('@stremio/stremio-icons/react');
-const {
-    Button,
-    Image,
-    useProfile,
-    platform,
-    useToast,
-    Popup,
-    useBinaryState,
-} = require('stremio/common');
+const { Button, Image, useProfile, platform, useToast, Popup, useBinaryState } = require('stremio/common');
 const { useServices } = require('stremio/services');
 const StreamPlaceholder = require('./StreamPlaceholder');
 const { t } = require('i18next');
 
 const styles = require('./styles');
 
-const Stream = ({
-    className,
-    videoId,
-    videoReleased,
-    addonName,
-    name,
-    description,
-    thumbnail,
-    progress,
-    deepLinks,
-    ...props
-}) => {
+const Stream = ({ className, videoId, videoReleased, addonName, name, description, thumbnail, progress, deepLinks, ...props }) => {
     const profile = useProfile();
     const toast = useToast();
     const { core } = useServices();
@@ -69,27 +50,36 @@ const Stream = ({
     }, []);
 
     const href = React.useMemo(() => {
-        return deepLinks
-            ? deepLinks.externalPlayer
-                ? deepLinks.externalPlayer.web
-                    ? deepLinks.externalPlayer.web
-                    : deepLinks.externalPlayer.openPlayer
-                        ? deepLinks.externalPlayer.openPlayer[platform.name]
-                            ? deepLinks.externalPlayer.openPlayer[platform.name]
-                            : deepLinks.externalPlayer.playlist
-                        : deepLinks.player
-                : deepLinks.player
-            : null;
+        return deepLinks ?
+            deepLinks.externalPlayer ?
+                deepLinks.externalPlayer.web ?
+                    deepLinks.externalPlayer.web
+                    :
+                    deepLinks.externalPlayer.openPlayer ?
+                        deepLinks.externalPlayer.openPlayer[platform.name] ?
+                            deepLinks.externalPlayer.openPlayer[platform.name]
+                            :
+                            deepLinks.externalPlayer.playlist
+                        :
+                        deepLinks.player
+                :
+                deepLinks.player
+            :
+            null;
     }, [deepLinks]);
 
     const download = React.useMemo(() => {
-        return href === deepLinks?.externalPlayer?.playlist
-            ? deepLinks.externalPlayer.fileName
-            : null;
+        return href === deepLinks?.externalPlayer?.playlist ?
+            deepLinks.externalPlayer.fileName
+            :
+            null;
     }, [href, deepLinks]);
 
     const target = React.useMemo(() => {
-        return href === deepLinks?.externalPlayer?.web ? '_blank' : null;
+        return href === deepLinks?.externalPlayer?.web ?
+            '_blank'
+            :
+            null;
     }, [href, deepLinks]);
 
     const markVideoAsWatched = React.useCallback(() => {
@@ -98,28 +88,27 @@ const Stream = ({
                 action: 'MetaDetails',
                 args: {
                     action: 'MarkVideoAsWatched',
-                    args: [{ id: videoId, released: videoReleased }, true],
-                },
+                    args: [{ id: videoId, released: videoReleased }, true]
+                }
             });
         }
     }, [videoId, videoReleased]);
 
-    const onClick = React.useCallback(
-        (event) => {
-            if (profile.settings.playerType !== null) {
-                markVideoAsWatched();
-                toast.show({
-                    type: 'success',
-                    title: 'Stream opened in external player',
-                    timeout: 4000,
-                });
-            }
+    const onClick = React.useCallback((event) => {
+        if (profile.settings.playerType !== null) {
+            markVideoAsWatched();
+            toast.show({
+                type: 'success',
+                title: 'Stream opened in external player',
+                timeout: 4000
+            });
+        }
 
-            if (typeof props.onClick === 'function') {
-                props.onClick(event);
-            }
-        },
-        [props.onClick, profile.settings, markVideoAsWatched]
+        if (typeof props.onClick === 'function') {
+            props.onClick(event);
+        }
+    },
+    [props.onClick, profile.settings, markVideoAsWatched]
     );
 
     const copyMagneticLinkToClipboard = React.useCallback((event) => {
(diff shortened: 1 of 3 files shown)


tymmesyde:
Hi! Thank you for your PR,
Are you able to make it a open a context menu on right click with an option for copying it?
If it's too big of a task, we can continue development of this feature based on your branch

Similar to how it's done when right clicking a Video component:
image

Here is the code for it: https://github.com/Stremio/stremio-web/blob/development/src/routes/MetaDetails/VideosList/Video/Video.js#L163

Namyts:
@tymmesyde - Added the context menu. Changed the Steam to use Popup component (like Video). Menu has a Play, and Copy infohash button, but would need a translation string (`CTX_COPY_INFOHASH`)

src/routes/MetaDetails/StreamsList/Stream/styles.less
```diff
@@ -105,6 +113,33 @@
     }
 }
 
+.context-menu-container {
```
tymmesyde: This should be inside `.stream-container` otherwise it applies to every `.context-menu-container` elements

src/routes/MetaDetails/StreamsList/Stream/styles.less
```diff
@@ -128,3 +163,29 @@
         }
     }
 }
+
+@media only screen and (max-width: @minimum) {
+    .video-container {
```
tymmesyde: `.video-container` should be `.stream-container`

src/routes/MetaDetails/StreamsList/StreamsList.js
```diff
@@ -64,96 +64,106 @@ const StreamsList = ({ className, video, ...props }) => {
...
         };
     }, [streamsByAddon, selectedAddon]);
     return (
         <div className={classnames(className, styles['streams-list-container'])}>
-            {
-                props.streams.length === 0 ?
-                    <div className={styles['message-container']}>
-                        <Image className={styles['image']} src={require('/images/empty.png')} alt={' '} />
-                        <div className={styles['label']}>No addons were requested for streams!</div>
+            {props.streams.length === 0 ? (
```
tymmesyde: It seems like your IDE did automatic indentation and formatting in lots of place, can you revert those?

tymmesyde (changes requested):
Thank you, there are some things that needs to be addressed

src/routes/MetaDetails/StreamsList/StreamsList.js
```diff
@@ -64,96 +64,106 @@ const StreamsList = ({ className, video, ...props }) => {
...
+                                videoId={video?.id}
+                                videoReleased={video?.released}
+                                addonName={stream.addonName}
+                                name={stream.name}
+                                description={stream.description}
+                                thumbnail={stream.thumbnail}
+                                progress={stream.progress}
+                                deepLinks={stream.deepLinks}
+                                onClick={stream.onClick}
+                                infoHash={stream.infoHash}
```
tymmesyde: Instead of passing the stream infoHash to the Stream component
It could use the `deepLinks.externalPlayer.download` inside the Stream component, this ways it would return the full magnet link when available and also allow to copy the url of other streams

Namyts:
@tymmesyde - Oops, had Prettier running! Reverted the weird formatting to clean up merge diffs. Also switched to copy magnetic link. This also removes need for translation since it uses `t('CTX_COPY_VIDEO_DOWNLOAD_LINK')` and `t('PLAYER_COPY_DOWNLOAD_LINK_SUCCESS')` . I assume these will start working correctly at some point, since I am not seeing the correct values locally...

src/routes/MetaDetails/StreamsList/Stream/Stream.js
```diff
@@ -74,41 +108,107 @@ const Stream = ({ className, videoId, videoReleased, addonName, name, descriptio
...
+    );
+
+    const renderMenu = function renderMenu() {
+        return (
+            <div className={styles['context-menu-content']} onPointerDown={popupMenuOnPointerDown} onContextMenu={popupMenuOnContextMenu} onClick={popupMenuOnClick} onKeyDown={popupMenuOnKeyDown}>
+                <Button className={styles['context-menu-option-container']} title={t('CTX_PLAY')} onClick={onClick}>
+                    <div className={styles['context-menu-option-label']}>{t('CTX_PLAY')}</div>
+                </Button>
+                {deepLinks?.externalPlayer?.download && <Button className={styles['context-menu-option-container']} title={t('CTX_COPY_VIDEO_DOWNLOAD_LINK')} onClick={copyMagneticLinkToClipboard}>
+                    <div className={styles['context-menu-option-label']}>{t('CTX_COPY_VIDEO_DOWNLOAD_LINK')}</div>
```
tymmesyde, line 191: `CTX_COPY_STREAM_LINK` can be used instead of `CTX_COPY_VIDEO_DOWNLOAD_LINK`
https://github.com/Stremio/stremio-translations/blob/master/en-US.json#L765

tymmesyde: CTX_COPY_STREAM_LINK can be used instead of CTX_COPY_VIDEO_DOWNLOAD_LINK
https://github.com/Stremio/stremio-translations/blob/master/en-US.json#L765

src/routes/MetaDetails/StreamsList/Stream/Stream.js
```diff
@@ -74,41 +108,107 @@ const Stream = ({ className, videoId, videoReleased, addonName, name, descriptio
...
     }, [props.onClick, profile.settings, markVideoAsWatched]);
 
+    const copyMagneticLinkToClipboard = React.useCallback((event) => {
+        event.preventDefault();
+        if (deepLinks?.externalPlayer?.download && navigator?.clipboard) {
+            navigator.clipboard.writeText(deepLinks.externalPlayer.download)
+                .then(() => {
+                    toast.show({
+                        type: 'success',
+                        title: t('PLAYER_COPY_DOWNLOAD_LINK_SUCCESS'),
```
tymmesyde: `PLAYER_COPY_STREAM_SUCCESS` can be used instead of `PLAYER_COPY_DOWNLOAD_LINK_SUCCESS`
https://github.com/Stremio/stremio-translations/blob/master/en-US.json#L188

src/routes/MetaDetails/StreamsList/Stream/Stream.js
```diff
@@ -74,41 +108,107 @@ const Stream = ({ className, videoId, videoReleased, addonName, name, descriptio
...
+                        timeout: 4000
+                    });
+                })
+                .catch(() => {
+                    toast.show({
+                        type: 'error',
+                        title: t('PLAYER_COPY_DOWNLOAD_LINK_ERROR'),
```
tymmesyde: `PLAYER_COPY_STREAM_ERROR` can be used instead of `PLAYER_COPY_DOWNLOAD_LINK_ERROR`
https://github.com/Stremio/stremio-translations/blob/master/en-US.json#L189

src/routes/MetaDetails/StreamsList/Stream/Stream.js
```diff
@@ -74,41 +108,107 @@ const Stream = ({ className, videoId, videoReleased, addonName, name, descriptio
...
-                        <div className={styles['thumbnail-container']} title={name || addonName}>
-                            <Image
-                                className={styles['thumbnail']}
-                                src={thumbnail}
-                                alt={' '}
-                                renderFallback={renderThumbnailFallback}
-                            />
+    const renderLabel = React.useMemo(
+        () =>
+            function renderLabel({ className, thumbnail, progress, children, ...props }) {
```
tymmesyde:
```suggestion
    const renderLabel = React.useMemo(() => function renderLabel({ className, addonName, name, description, thumbnail, progress, children, ...props }) {
```
Here it needs to have the `addonsName`, `name` and `description` too, otherwise they don't update when using the addon selector/filter

src/routes/MetaDetails/StreamsList/Stream/Stream.js
```diff
@@ -74,41 +108,107 @@ const Stream = ({ className, videoId, videoReleased, addonName, name, descriptio
...
-            <div className={styles['description-container']} title={description}>{description}</div>
-            <Icon className={styles['icon']} name={'play'} />
-        </Button>
+        );
+    };
+
+    return (
+        <Popup
+            className={className}
+            thumbnail={thumbnail}
```
tymmesyde, line 215:
```suggestion
            thumbnail={thumbnail}
            addonName={addonName}
            name={name}
            description={description}
```
Needed for the `renderLabel` function

src/routes/MetaDetails/StreamsList/Stream/Stream.js
```diff
@@ -74,41 +108,107 @@ const Stream = ({ className, videoId, videoReleased, addonName, name, descriptio
...
+                {deepLinks?.externalPlayer?.download && <Button className={styles['context-menu-option-container']} title={t('CTX_COPY_VIDEO_DOWNLOAD_LINK')} onClick={copyMagneticLinkToClipboard}>
+                    <div className={styles['context-menu-option-label']}>{t('CTX_COPY_VIDEO_DOWNLOAD_LINK')}</div>
+                </Button>}
             </div>
-            <div className={styles['description-container']} title={description}>{description}</div>
-            <Icon className={styles['icon']} name={'play'} />
-        </Button>
+        );
+    };
+
```
tymmesyde, line 196: This can be added to close the menu automatically when the route change (like when clicking Play)
```suggestion
    React.useEffect(() => {
        if (!routeFocused) {
            closeMenu();
        }
    }, [routeFocused]);
```

`routeFocused`  needs to be declared, just like in the Video component:
```
const routeFocused = useRouteFocused();
```


After this, you can remove the `onClick` event of the `CTX_PLAY` Button

tymmesyde (changes requested):
Thank you for your changes, no problem!
For the translations strings i added the ones that should be working and suggested some others changes

src/routes/MetaDetails/StreamsList/Stream/Stream.js
tymmesyde, line 211: This can be added to close the menu automatically when the route change (like when clicking Play)
```suggestion
    React.useEffect(() => {
        if (!routeFocused) {
            closeMenu();
        }
    }, [routeFocused]);
```

`routeFocused`  needs to be declared, just like in the Video component:
```
const routeFocused = useRouteFocused();
```


After this, you can remove the `onClick` event of the `CTX_PLAY` Button

src/routes/MetaDetails/StreamsList/Stream/Stream.js
Namyts: nice!

src/routes/MetaDetails/StreamsList/Stream/Stream.js
Namyts: Looks like it is working now :)

Namyts:
@tymmesyde - made the changes

src/routes/MetaDetails/StreamsList/Stream/Stream.js
```diff
@@ -74,41 +116,110 @@ const Stream = ({ className, videoId, videoReleased, addonName, name, descriptio
...
+        } else {
+            toast.show({
+                type: 'error',
+                title: t('PLAYER_COPY_STREAM_ERROR'),
+                timeout: 4000,
+            });
+        }
+        closeMenu();
+    }, []);
```
tymmesyde: Here it's missing the `deepLinks` dep for reactivity

src/routes/MetaDetails/StreamsList/Stream/Stream.js
```diff
@@ -74,41 +116,110 @@ const Stream = ({ className, videoId, videoReleased, addonName, name, descriptio
...
+                </Button>
+                {deepLinks?.externalPlayer?.download && <Button className={styles['context-menu-option-container']} title={t('CTX_COPY_STREAM_LINK')} onClick={copyMagneticLinkToClipboard}>
+                    <div className={styles['context-menu-option-label']}>{t('CTX_COPY_STREAM_LINK')}</div>
+                </Button>}
             </div>
-            <div className={styles['description-container']} title={description}>{description}</div>
-            <Icon className={styles['icon']} name={'play'} />
-        </Button>
+        );
+    };
```
tymmesyde: `renderMenu` it should use `useMemo` with `deepLinks` and `copyMangeticLinkToClipboard` as deps

src/routes/MetaDetails/StreamsList/Stream/Stream.js
```diff
@@ -74,41 +116,110 @@ const Stream = ({ className, videoId, videoReleased, addonName, name, descriptio
         }
     }, [props.onClick, profile.settings, markVideoAsWatched]);
 
+    const copyMagneticLinkToClipboard = React.useCallback((event) => {
```
tymmesyde: It can be just named `copyStreamLink` now that we use `deepLinks.externalPlayer.download`

tymmesyde (changes requested):
Thanks, last changes needed
I can push commits to your branch to finish it up if you prefer

tymmesyde:
You can also create and use this memo instead of writing the full `deepLinks.externalPlayer.download` where it's needed
```js
const streamLink = React.useMemo(() => {
    return deepLinks?.externalPlayer?.download;
}, [deepLinks]);
 ```

src/routes/MetaDetails/StreamsList/Stream/Stream.js
Namyts: Might not need deepLinks for this one? (deepLinks would change streamLink would change copyStreamLink would change renderMenu)

Namyts:
My bad for missing the memo opportunities! I've always let React continue its inefficiencies unless it was critical 😅 React's new compiler should especially help with all of this! 

I also try to avoid nested component definitions. Makes it more confusing what values to pass as props vs what values to take from the parent. Probably harder for optimisation too, which is why these memos are even more important in these files... good spot

tymmesyde (approved):

tymmesyde:
> My bad for missing the memo opportunities! I've always let React continue its inefficiencies unless it was critical 😅 React's new compiler should especially help with all of this!
> 
> I also try to avoid nested component definitions. Makes it more confusing what values to pass as props vs what values to take from the parent. Probably harder for optimisation too, which is why these memos are even more important in these files... good spot

No problem! Yeah to be fair this menu implementation is a bit hard to understand

kKaskak:
Mostly LGTM 
There is one edge case worth mentioning. When there are only 1-2 streams being displayed, it's hard to access the menu. Can we fix this?

tymmesyde:
> Mostly LGTM There is one edge case worth mentioning. When there are only 1-2 streams being displayed, it's hard to access the menu. Can we fix this? 

I don't think this is an issue related to this PR, it happens too with videos since there is also a context menu there
Opened a PR to fix it: https://github.com/Stremio/stremio-web/pull/665

kKaskak (approved):

PR #661, opened 2024-07-22
feat: Introduce IMDb Rating on Card View

anant-j:
Hi, 

As an iOS user (web.stremio.com on PWA), I would love to view the IMDb ratings on the home screen along with the movie title. 
This simple change introduces the rating next to the title.
Note:
- Only enabled on the Home/Board Tab, as Discovery tab already has this in preview side pane.
- Disabled on small screen/mobile devices, as title gets shortened and illegible when this is introduced on small screens.

Before I continue developing this further, just want to check if this is something that could be considered for introducing to the platform? I am aware that there is RPDB but that is a paid option, and this IMDb rating is already present in current props.

If this sounds like a feasible idea, here are the further action items (in priority) that could be worked on before shipping:
- [x] Introduce the slot for rating next to title
- [ ] Allow user to turn this feature on/off from settings
- [ ] Allow user to place single slot on left or right of title
- [ ] Allow user to select between runtime/release year/IMDb rating to display in the slot OR display all 3 just above the title
- [ ] Consider supporting Rotten Tomatoes/TMDB/Metacritic integration in the future

Feature Screenshots:
| Feature | Production |
|---------|------------|
| image | image |

### DIFF
--- a/src/common/MetaItem/MetaItem.js
+++ b/src/common/MetaItem/MetaItem.js
@@ -211,7 +211,7 @@ MetaItem.propTypes = {
         category: PropTypes.string,
         name: PropTypes.string,
         url: PropTypes.string
-    })),
+    }))
 };
 
 module.exports = MetaItem;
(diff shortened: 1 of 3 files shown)


kKaskak:
we are sorry but design team rejected this. Thank you for your submission anyways

PR #662, opened 2024-07-22
feat: Introduce Hide/Show option for each row

anant-j:
Introducing a small change that adds the option to hide unwanted rows during a browsing session on Stremio.
Sometimes for users with too many add-ons, it becomes a chore to scroll through unwanted rows of content that the user does not want. A simple solution to this is to temporarily hide the content-row and give user the control.

Feature demo:
ezgif-6-02fef43591

Other related features/improvements that may be considered in the future:
- [x] Ability to show/hide rows and persist during session
- [ ] Ability to permanently hide row, this may be combined with the ability to re-order addons as per user's preference
- [ ] Store state and persist outside session (maybe as a setting stored on user profile) and sync across all devices
- [ ] Add animations to provider a smoother user experience.

### DIFF
--- a/src/common/MetaRow/MetaRow.js
+++ b/src/common/MetaRow/MetaRow.js
@@ -13,6 +13,7 @@ const styles = require('./styles');
 
 const MetaRow = ({ className, title, catalog, message, itemComponent, notifications }) => {
     const t = useTranslate();
+    const [isHidden, setIsHidden] = React.useState(false);
 
     const catalogTitle = React.useMemo(() => {
         return title ?? t.catalogTitle(catalog);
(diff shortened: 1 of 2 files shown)


anant-j:
Some requests asking for this feature:
- https://github.com/Stremio/stremio-features/issues/567
- https://github.com/Stremio/stremio-features/issues/661
- https://github.com/Stremio/stremio-features/issues/663

anant-j:
Hi @tymmesyde, possible to get a review on this?

nalabdou:
Hi @anant-j,

Thank you for your effort, but the mentioned issues are asking for something slightly different. They are looking for a full layout customizer, which includes features like permanently hiding sections or re-sorting them—not just front-end adjustments. I believe achieving that would require more significant changes to the files and backend logic.

Best regards,
Nadim.

kKaskak:
dt rejected

PR #663, opened 2024-07-22
fix: Streaming server settings - use empty string instead of null for remoteHttps

elpiel:
Adds the UI fix for the core's change to use empty string instead of `null`. This is in-par with Desktop and it will now use the same options on web and Desktop.

Needs core: https://github.com/Stremio/stremio-core/pull/702


kKaskak (approved):

tymmesyde:
It needs to be changed in server and desktop because we shouldn't use and empty string as `null`

elpiel:
# TODOs

- [x] Needs core bump

elpiel:
Ready for merging.

PR #696, opened 2024-09-30
Fix for not being able to scroll seasons dropdown

arpitbatra123:
Before: 

Seasons dropdown is clipped, can't scroll
localhost_8080_

After:
Can scroll now
localhost_8080_ (1)
On mobile
localhost_8080_ (2)

### DIFF
--- a/src/routes/MetaDetails/VideosList/SeasonsBar/styles.less
+++ b/src/routes/MetaDetails/VideosList/SeasonsBar/styles.less
@@ -80,6 +80,11 @@
             max-height: calc(3.2rem * 7);
             overflow: auto;
         }
+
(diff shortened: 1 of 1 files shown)


kKaskak:
Hi. Could you instead modify the `open` state styles in 
`src/common/MultiselectMenu/Dropdown/Dropdown.less` ?
This place would be more fit for the change. 

You can do something like this ⇾
```
&.open {
   display: block;
   max-height: calc(3.2rem * 10);
   overflow: auto;
}
```

And also in the same file you can add the mobile styles too

Thanks for your contribution

kKaskak:
https://github.com/Stremio/stremio-web/pull/697

PR #727, opened 2024-11-16
fix: modal dialog styles issues

### LINKED ISSUE #721: UX: Long manifest URLs prevent add-on installation button from displaying
### Description
The addon installation button becomes inaccessible when the manifest URL is extremely long, causing UI rendering issues in the web interface.

### Current Behavior
- Long manifest URLs occupy the entire width of the installation dialog
- The installation button is pushed out of view
- Users cannot proceed with installation through normal means

### Workaround
Currently, users must:
1. Open browser developer tools
2. Modify the URL text element
3. Remove/shorten the URL to make the install button visible

### Expected Behavior
- The manifest URL should be truncated in the UI
- Consider showing only the first portion of the URL (e.g., `https://example.com/...`)
- Installation button should always be visible and accessible

#### Issue: Hidden Install Button
Hidden Install Button

#### Cleanup URL
Cleanup URL

kKaskak:
Fixes:

- [x] Content wasn't accessible when the addon info is too long
- [x] Button text wasn't visible on mobile because of mistake in class placement

### DIFF
--- a/src/common/ModalDialog/ModalDialog.js
+++ b/src/common/ModalDialog/ModalDialog.js
@@ -70,7 +70,7 @@ const ModalDialog = ({ className, title, buttons, children, dataset, onCloseRequ
                             :
                             null
                     }
-                    <div className={styles['modal-dialog-content']}>
+                    <div className={styles['body-container']}>
                         {children}
                     </div>
                     {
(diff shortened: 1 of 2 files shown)


tymmesyde:
Pushed a commit to rename `modal-dialog-content` to `body-content` because it was using the same name as the container

tymmesyde (approved):

PR #728, opened 2024-11-20
Added possibility to set HTTP_PORT with environmental variable

tubbadu:
default value is still 8080, but if the env variable is set that one will be used

### DIFF
--- a/http_server.js
+++ b/http_server.js
@@ -4,7 +4,7 @@
 
 const INDEX_CACHE = 7200;
 const ASSETS_CACHE = 2629744;
-const HTTP_PORT = 8080;
+const HTTP_PORT = process.env.HTTP_PORT ? process.env.HTTP_PORT : 8080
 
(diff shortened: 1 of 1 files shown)


http_server.js
```diff
@@ -4,7 +4,7 @@
 
 const INDEX_CACHE = 7200;
 const ASSETS_CACHE = 2629744;
-const HTTP_PORT = 8080;
+const HTTP_PORT = process.env.HTTP_PORT ? process.env.HTTP_PORT : 8080
```
Adig0, line 7: `const HTTP_PORT = process.env.HTTP_PORT ?? 8080`

Adig0, line 7:
```suggestion
const HTTP_PORT = process.env.HTTP_PORT ?? 8080
```

kKaskak:
Hi, Better for the repo to not have any - even optional - ENVs.

PR #731, opened 2024-11-27
Fix: styles on metaDetails / Streams page

kKaskak:
## Fixes:

- The addon label was cut if the addon name was too long on the `Stream` component 
- Long addon names were casuing overflow on the `multiSelect` component

### Tested on: 
- [x]  iPad PWA

### DIFF
--- a/src/routes/MetaDetails/StreamsList/Stream/styles.less
+++ b/src/routes/MetaDetails/StreamsList/Stream/styles.less
@@ -58,7 +58,7 @@
 
         .addon-name {
             width: 7rem;
-            max-height: 3.6em;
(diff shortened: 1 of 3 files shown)


src/routes/MetaDetails/StreamsList/Stream/styles.less
```diff
@@ -58,7 +58,7 @@
 
         .addon-name {
             width: 7rem;
-            max-height: 3.6em;
+            max-height:4em;
```
tymmesyde: best to just remove the max-height
```suggestion
```

tymmesyde (approved):

PR #732, opened 2024-11-28
fix: navbar item label was cut

kKaskak:


TRtomasz:
if it closes any stremio-task pls link next time so it is easier to check what actually gets fixed

kKaskak:
> if it closes any stremio-task pls link next time so it is easier to check what actually gets fixed

This wasn't closing any tasks or issues I believe. 
This was a problem with the calendar item Label on the Vertical Navbar component 
And it was specifically related to the calendar PR

PR #740, opened 2024-12-11
fix: nav-menu height

kKaskak:
image

- added some empty space so the menu ends before the end of the page

### Before
image


TRtomasz:
this is fix for small displays right? lgtm

kKaskak:
> this is fix for small displays right? lgtm

yeah just to make some spacing from the bottom so it doesnt look like its overflowing

PR #742, opened 2024-12-11
fix: multiselectMenu and Multiselect height

kKaskak:
- The height of the `Multiselect` and `MultiselectMenu` was not calculated properly.
- After fixing the problem with the pop-up overflow some time ago a new problem is getting fixed here with dropdowns inside pop-ups on being visible (another time overflow issue) i propose we move away from popups completely in the future

-  Changes  introduced: 
 1. We set the height only globally for the component and try not to set it locally for each case.
 2. We use `@media (orientation: landscape) and (max-width: @xsmall)` to set the size for smaller landscape devices (there are no devices that are vertical and are smaller than `calc(3.2rem * 7)`)
 4. Overflow issue on popups having dropdowns resolved. 

- Future notes: 
 1. Move completely to using `MultiselectMenu` component instead of the old `MultiSelect`
### Now
image
image
image

### Before
image
image

### DIFF
--- a/src/routes/Discover/styles.less
+++ b/src/routes/Discover/styles.less
@@ -57,7 +57,6 @@
                     }
 
                     .multiselect-menu-container {
-                        max-height: calc(3.2rem * 7);
                         overflow: auto;
                     }
                 }
@@ -220,7 +219,7 @@
                 }
 
                 .multiselect-menu-container {
-                    max-height: calc(3.2rem * 4);
+                    max-height: calc(3.2rem * 3);
                     overflow: auto;
                 }
             }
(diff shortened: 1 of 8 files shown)


src/routes/Discover/styles.less
```diff
@@ -220,7 +219,7 @@
                 }
 
                 .multiselect-menu-container {
-                    max-height: calc(3.2rem * 4);
+                    max-height: calc(3.2rem * 3);
```
tymmesyde, line 222:
```suggestion
```

tymmesyde (approved):

PR #747, opened 2024-12-15
Fix search query handling to support multi-word queries

iAbhishekPanwar:
Changes Made:
Updated the URL construction logic to ensure the search query is captured accurately, including special characters and spaces. Ensured compatibility with all valid search query formats, avoiding malformed URLs.

Impact of the Fix:
Multi-word queries now work as expected, returning accurate and complete results. Improves the robustness of the URL handling logic.

Example:
Before the fix:

Searching "Alexa & Katie" would return results only for "Alexa."

After the fix:

The full query "Alexa & Katie" is correctly parsed, and results are returned for both "Alexa" and "Katie."

### DIFF
--- a/src/App/SearchParamsHandler.js
+++ b/src/App/SearchParamsHandler.js
@@ -14,7 +14,10 @@ const SearchParamsHandler = () => {
 
     const onLocationChange = () => {
         const { origin, hash, search } = window.location;
-        const { searchParams } = new URL(`${origin}${hash.replace('#', '')}${search}`);
+        // Fix: Ensure the entire query string is captured as a single string, including special characters like '&'
+        const url = new URL(`${origin}${hash.replace('#', '')}${search ? search : ''}`);
+        const { searchParams } = url;
+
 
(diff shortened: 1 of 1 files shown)


kKaskak:
Hey. Thanks for the contribution, however the results were always correctly displayed. The issue was in the `SearchBar` component here 
https://github.com/Stremio/stremio-web/blob/development/src/common/NavBar/HorizontalNavBar/SearchBar/SearchBar.js#L64

where it should have been encoded with `encodeURIcomponent` function like this 

```
const searchValue = `/search?search=${encodeURIComponent(event.target.value)}`;
```

if you want you can make this change instead so we can merge your PR.

iAbhishekPanwar:
Thanks for the reply, i'll make the change right away

iAbhishekPanwar:
I’ve made the changes and opened a new PR with the fix.

kKaskak:
> I’ve made the changes and opened a new PR with the fix.

Thank you will be reviewing the new PR then. 💯

PR #754, opened 2024-12-16
fix(BottomSheet): was hidden on ipad pro in portait mode

tymmesyde:


kKaskak (approved):
LGTM tested on ipad pro ✅

PR #765, opened 2024-12-21
refactor(styles): improve PWA styles on mobile devices (1)

kKaskak:
### First PR in the series of PRs adressing the big @IEduStu PR.

Reference for changes:
- Fixed html height / width in PWAs
- Fixed offsets and safe-areas issues
- https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-touch-callout
- Removed `-webkit-tap-highlight-color: transparent;` since it's applied on `html` globally in `styles.less` on all components
- Removed used `px` instead using `rem`
- Removed inline var declarations, using global vars from `app.less`
- Removed a variable that was basically a repetition of another variable; reusing instead
- Improved conditionals with `useMemo`
- Created new variables for some of the cases


kKaskak:
commit `b15b7bbb12e6efe53ce8471eaff0ef5dfdd6999c` is experimental

TRtomasz:
seems to be not affecting the new shell/web app

tymmesyde (dismissed):

kKaskak:
@IEduStu we will continue to use your commits by cherry picking to improve the Stremio App. Thank you for your contributions. 💯

PR #769, opened 2024-12-23
Fix Poster Sharpness

jaruba:
Chrome (on Windows only) was showing the posters with high sharpness, this made it hard to read text on some posters (including the title), this one css style seems to fix that and I didn't notice it impacting other cases yet.

### DIFF
--- a/src/common/MetaItem/styles.less
+++ b/src/common/MetaItem/styles.less
@@ -166,6 +166,7 @@
                 object-position: center;
                 object-fit: cover;
                 opacity: 0.9;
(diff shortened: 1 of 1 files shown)


kKaskak (approved):
for reference: https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-clip-margin
unable to test on windows
tested on Safari ✅

jaruba:
I'm merging, I tested on chrome for macos too, it seems fine and i think it may be more of a fix tbh

PR #813, opened 2025-01-29
Player: add context menu

Botsy:


src/components/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+    useEffect(() => {
+        on.forEach((ref) => ref.current && ref.current.addEventListener('contextmenu', onContextMenu));
+        return () => on.forEach((ref) => ref.current && ref.current.removeEventListener('contextmenu', onContextMenu));
+    }, [on]);
+
+    return active && createPortal((
+        <div
+            className={styles['context-menu-container']}
+            onMouseDown={close}
```
elpiel, line 84: Maybe we can add Escape for key to close the context menu as well.
Just need to be careful if the Player is on full screen to close the context menu first instead of exiting the full screen

elpiel (dismissed):
LGTM.
I've added one suggestion for the context menu closing.

kKaskak (approved):

PR #817, opened 2025-01-29
App(Shell): add updater banner

tymmesyde:
Closes https://github.com/Stremio/stremio-tasks/issues/357

Need https://github.com/Stremio/stremio-translations/pull/857


TRtomasz (dismissed):
tested with `--force-update` seems to be fine

PR #824, opened 2025-02-05
App: support video hardware decoding

tymmesyde:
Closes https://github.com/Stremio/stremio-tasks/issues/377

Need https://github.com/Stremio/stremio-video/pull/97


TRtomasz (dismissed):
switch is working, looks like shell does not support it yet cause after turning it off it still uses hw dec

PR #833, opened 2025-02-13
Calendar: Fix items scroll in view to ensure borders remain visible

ivelinmegdanov:
This PR fixes an issue where calendar list items were partially hidden when scrolled into view. The issue was caused by how some browsers handles `scrollIntoView({ block: 'start' })`, which sometimes ignores borders and padding.

Used scroll margin/padding block at the start

Closes [Issue #372](https://github.com/Stremio/stremio-tasks/issues/372)

### DIFF
--- a/src/routes/Calendar/List/Item/Item.less
+++ b/src/routes/Calendar/List/Item/Item.less
@@ -7,8 +7,6 @@
     flex-direction: column;
     background-color: var(--overlay-color);
     border-radius: var(--border-radius);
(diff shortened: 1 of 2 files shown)


kKaskak (changes requested):
instead use `scroll-margin-block-start` combined with `scroll-padding-block-start`

kKaskak (dismissed):

tymmesyde (approved):

PR #836, opened 2025-02-17
Discover: Remove duplicated filters in filter modal

Botsy:
This fix for Discover page shows in the filters modal only the filters that are not visible as a dropdown above the results. Also fixes the spacing issue of the filter button on larger screens with more than 3 filters.


src/components/Multiselect/Multiselect.js
```diff
@@ -10,16 +10,16 @@ const ModalDialog = require('stremio/components/ModalDialog');
...
-        return Array.isArray(props.options) ?
-            props.options.filter((option) => {
+    const filteredOptions = React.useMemo(() => {
+        return Array.isArray(options) ?
+            options.filter((option) => {
                 return option && (typeof option.value === 'string' || option.value === null);
             })
             :
             [];
-    }, [props.options]);
```
Botsy, line 22: I destructured the options from the props because they were being unnecessarily rendered in the html as props in the elements.

kKaskak (approved):

kKaskak (approved):

PR #855, opened 2025-02-28
Аpp: Replace Multiselect usage with MultiselectMenu

Botsy:


kKaskak (changes requested):
styles need to be preserved from before the changes + handle the case where we dont have a set language and default to english

kKaskak (approved):

PR #862, opened 2025-03-10
Discover Page: Fix infinite scroll issue on large screens

ivelinmegdanov:
Detects when there is too little content to trigger scrolling and 
automatically loads the next page if needed.

Closes [Issue #374](https://github.com/Stremio/stremio-tasks/issues/374)


kKaskak (changes requested):
still does not work on some larger displays

kKaskak (approved):

kKaskak (approved):

PR #871, opened 2025-03-20
App(Shell): support opening addons from deeplink

tymmesyde:


core1024 (approved):

TRtomasz:
i checked this now and it works only for the addons every other deeplink is still not working

tymmesyde:
> i checked this now and it works only for the addons every other deeplink is still not working

yes, as the title says it's only for addons

PR #876, opened 2025-03-24
App: ignore apple well known folder for gh

kKaskak:


elpiel (changes requested):
For this type of ignored files it's good to use a global gitignore instead of adding this to every project.

`~/.gitconfig`:
```
[core]
	excludesfile = /home/USER/.gitignore_global
```

PR #894, opened 2025-04-15
(Subtitles): Shows subtitle id on hover instead of url

ivelinmegdanov:
Requires: https://github.com/Stremio/stremio-core/pull/804

Closes [Issue #403](https://github.com/Stremio/stremio-tasks/issues/403)


kKaskak:
outdated

Issue #801, opened 2025-01-17
[Feature]: Open Streams in External Player by default


hruzgar:
### Feature Description

**How it is currently:**
Currently all streams are opened to the stremio player by default. After it opens you can then select VLC or other players from the menu.

**My proposed solution:**
You should be able to select the external player you want in the settings to open streams by default in that player. So that you don't have to open the stremio player first every time.

### Additional Context or Screenshots

This is what you currently have to do if you want to play the stream in an external player:

![Image](https://github.com/user-attachments/assets/3bb083d5-d499-47dc-aa2e-a4d74d11f6f1)

This is where you should be able to select the external player you want and streams then open directly in that player without you having to open the stremio player first:

![Image](https://github.com/user-attachments/assets/64a86494-fec1-4088-aac4-87e8fd886a5e)

### Code of Conduct

- [x] I agree

ramirocardelli:
Hi! Could you please assign this issue to me? I'd love to work on it.
Let me know if there are any guidelines I should follow before starting.

arabyalhomsi:
Hello, did you manage to merge this with main? need this bad

ramirocardelli:
> Hello, did you manage to merge this with main? need this bad

Hello, I did not manage to merge, all checks are passed but there's still a review required

mythbuster74:
What is the status of this?  I'd love this feature

depleur:
Hey, any updates with this?

stickyburn:
Its been closed as 'not planned', which is weird if a proper PR was already submitted by a contributor 🙃

depleur:
@stickyburn I think this was merged and pushed because I now see this setting as an option:

<img width="596" height="239" alt="Image" src="https://github.com/user-attachments/assets/cf060fed-c36c-4c46-92fe-4c865602b870" />

stickyburn:
@depleur huh.. I dont see it on my mac, version 4.4.172

anilabhadatta:
@ramirocardelli how to add Energy media player here?
Is there any way to modify server.js to add the option here?

PR #899, opened 2025-04-28
Feature: Allow streams to open in external player by default (#801)

ramirocardelli:
As is states in the issue, I added the option of VLC on "Play in external player"
Also when "play" button is clicked, if VLC is the default option, it downloads the file and opens it on VLC automatically, if not, no behavior is changed


kKaskak:
Hi. Thanks for the PR. Please fix the build issues first.

eeriemyxi:
Could you please also make it work with MPV? Reading the changes, it seems like it's only for VLC users.

kKaskak:
after looking into that i am not sure what was the goal of the pr this is the behavior that we had before all streams are directly opened in external player there is no need to add this logic here?

PR #920, opened 2025-05-27
App: add missing translations

Botsy:
In this PR all hardcoded strings in web app are replaced with translation keys that were added in stremio-translations. A script for locating the hardcoded strings is added, but could be improved in future to detect more precisely all edge cases;

### DIFF
--- a/scan-i18n-report.js
+++ b/scan-i18n-report.js
@@ -1,0 +1,107 @@
+const fs = require('fs');
+const path = require('path');
+const recast = require('recast');
+const babelParser = require('@babel/parser');
+
+const directoryToScan = './src';
+const report = [];
+
+function toKey(str) {
+    return str
+        .toLowerCase()
+        .replace(/[^a-z0-9\s]/g, '')
+        .replace(/\s+/g, '_')
+        .slice(0, 40);
+}
+
+function scanFile(filePath) {
+    try {
+        const code = fs.readFileSync(filePath, 'utf8');
+        const ast = babelParser.parse(code, {
+            sourceType: 'module',
+            plugins: [
+                'jsx',
+                'typescript',
+                'classProperties',
+                'objectRestSpread',
+                'optionalChaining',
+                'nullishCoalescingOperator',
+            ],
+            errorRecovery: true,
+        });
+
+        recast.types.visit(ast, {
+            // Text directly inside JSX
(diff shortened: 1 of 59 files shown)


scan-i18n-report.js
tymmesyde, line 1: Can you convert this to a test instead without writing a csv file ?

src/common/useTranslate.js
```diff
@@ -9,7 +9,7 @@ const useTranslate = () => {
     const string = useCallback((key) => t(key), [t]);
 
     const stringWithPrefix = useCallback((value, prefix, fallback = null) => {
-        const key = `${prefix}${value}`;
+        const key = `${prefix || ''}${value}`;
```
tymmesyde: The point of this function is to use a prefix so it can't be an empty string

tests/i18nScan.test.js
```diff
@@ -0,0 +1,109 @@
...
+            console.log('🚨 Hardcoded strings found:');
+            report.forEach((row) => {
+                console.log(`File: ${row.file}, Line: ${row.line}, String: "${row.string}", Suggested Key: ${row.key}`);
+            });
+        }
+
+        // Optional: expect no hardcoded strings if you want the test to fail in that case
+        // expect(report.length).toBe(0);
+    });
+});
```
tymmesyde: Could we do something like this instead?
```suggestion
const report = [];
walk(directoryToScan, report);

describe.each(report)('hardcoded string', (entry) => {
    it(`should not have "${entry.string}" in ${entry.file} at line ${entry.line}`, () => {
        expect(entry.string).toBeFalsy();
    });
});
```

src/common/useTranslate.js
Botsy: You are right. It was used for discover extra dropdown options, but we don't need it, so I will revert this change and fix the options for extra dropdowns.

tests/i18nScan.test.js
Botsy: Wouldn't failing of those tests affect the build process?

tymmesyde: Yes, so we can enforce using translations strings when adding stuff to the codebase so we don't have to go back again like you did to add missing strings

tymmesyde (approved):

PR #935, opened 2025-06-12
Player: Fix side drawer layout issue

### LINKED ISSUE #934: [Bug]: Video player sideDrawer show description issue
### Stremio-Web Version

/development branch

### Browser

Firefox

### Platform / Device type

Windows

### What Happened?

# Summary

When playing a video and accessing the `sideDrawer`, If the show [description](https://github.com/Stremio/stremio-web/blob/d75c9b1d99906285899e41d1b051d8184ec505c3/src/routes/Player/SideDrawer/SideDrawer.tsx#L84) is too long, it makes the videos list smaller and harder to use.
# What I did

Navigate to watch a show, selected an episode, and I clicked on the `sideDrawer` to change the episode. 

# Expected to happen

Click on `sideDrawer` and scroll to the episode i want to watch

# Actually happened

It was hard to use because of the long description.
# Proposed solutions

1. Truncate the description when it reaches a certain length. Maybe make it expandable, or redirect the user to the show's main page to read the full description.

2. Don't show the show's description in the player's `sideDrawer`. I don't think the user needs to see the show description there every time they access the video player `sideDrawer`. 

### Logs

```shell

```

### Notes

App Version: 5.0.0-beta.23

Examples:

## 1
Image

## 2
Image

### Code of Conduct

- [x] I agree

val-makkas:
This PR implements a "Show more/less" button for long descriptions in the SideDrawer component.
When the description exceeds 175 characters, it is truncated and a button appears to expand/collapse the full text.

Closes #934

Please review and let me know if any changes are needed!

### DIFF
--- a/src/routes/Player/SideDrawer/SideDrawer.less
+++ b/src/routes/Player/SideDrawer/SideDrawer.less
@@ -67,6 +67,24 @@
         }
     }
 
+    .show-more-btn {
+        background: none;
+        border: none;
+        color: #fff;
+        cursor: pointer;
+        font: inherit;
+        padding: 0.1rem 1.5rem 0.2rem 1.5rem;
+        margin: 0.1rem auto 0.2rem auto;
+        display: block;
+        text-align: center;
+        transition: text-decoration 0.2s;
+        text-decoration: none;
+
+        &:hover {
+            text-decoration: underline;
+        }
+    }
+
     .series-content {
         flex: 2;
         display: flex;
--- a/src/routes/Player/SideDrawer/SideDrawer.less
+++ b/src/routes/Player/SideDrawer/SideDrawer.less
@@ -57,31 +57,13 @@
     .info {
         padding: @padding;
         overflow-y: auto;
-        flex: none;
+        flex: 1;
 
         .side-drawer-meta-preview {
             .action-buttons-container {
                 padding-top: 0;
                 margin-top: 0;
             }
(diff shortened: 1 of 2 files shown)


kKaskak:
### TODO

- [x] lower complexity, a better approach to this problem would be changing this line of code:
https://github.com/Stremio/stremio-web/pull/935/files#diff-5113a37954d2d1c497cc5c18419b298f2ec560a14f9aaa3d1b40f05282479931R60
to 
```
flex: 1;
```
simple = best

- [x] fix builds

val-makkas:
Thanks for the suggestion @kKaskak !

I did try replacing it with flex: 1, but it introduced an extra scrollbar inside the SideDrawer (so we end up with two components with scrollbars layered on top of each other). That made the UX a bit clunky so we can maybe try flex: 1 with the show more button too?

kKaskak:
> Thanks for the suggestion @kKaskak !
> 
> I did try replacing it with flex: 1, but it introduced an extra scrollbar inside the SideDrawer (so we end up with two components with scrollbars layered on top of each other). That made the UX a bit clunky so we can maybe try flex: 1 with the show more button too?

long description is an edge case, there is simply no need for an additional button there to cluster space.
Scrollable containers are not an issue. it is already like that on mobile as well. please check the to do list. thank you

kKaskak (approved):

PR #936, opened 2025-06-13
Player: Auto select current video in side drawer

mesalilac:
Adds a button in the `SideDrawer` of the video player to allow the user to jump to the currently playing video.

## Notes
- I used the compose icon because i couldn't find a better icon!

## Demo
Pi7_GIF_CMP


tymmesyde:
Hi, i think this should be the default behavior, so no need for a button

mesalilac:
> Hi, i think this should be the default behavior, so no need for a button

Good idea :+1:

mesalilac:
I'm facing a problem right now, the [`useEffect`](https://github.com/Stremio/stremio-web/pull/936/files#diff-083c71782a3716e7f1cb143b6c9bdd5a7ae217d5c7d5c49dca67ca29b22214a7R81) run takes a long time to `scrollIntoView` when the video list is long.
does anyone have a better way to do it?

> [!NOTE]
> The reason i have a `setTimeout` delay inside the `useEffect` is because `scrollIntoView` moves the sidebar elements if the sidebar transition is not complete.

kKaskak:
> I'm facing a problem right now, the [`useEffect`](https://github.com/Stremio/stremio-web/pull/936/files#diff-083c71782a3716e7f1cb143b6c9bdd5a7ae217d5c7d5c49dca67ca29b22214a7R81) run takes a long time to `scrollIntoView` when the video list is long. does anyone have a better way to do it?
> 
> Note
> 
> The reason i have a `setTimeout` delay inside the `useEffect` is because `scrollIntoView` moves the sidebar elements if the sidebar transition is not complete.

You can use `onTransitionEnd` to track when the transition ends and then fire the scroll. 

### TODO
- [x] use `onTransitionEnd` to eliminate the need for `setTimeout`
- [x] use `forwardRef` instead of creating a container wrapper for the `Video` component
- [x] clear prop naming

Thank you for the contribution

src/routes/Player/SideDrawer/SideDrawer.tsx
```diff
@@ -75,8 +76,21 @@ const SideDrawer = memo(forwardRef<HTMLDivElement, Props>(({ seriesInfo, classNa
         event.stopPropagation();
     };
 
+    const selectedVideoRef = useRef<HTMLDivElement>(null);
+
+    const jumpToPlayingNow = () => {
+        const { current } = selectedVideoRef;
+        if (!current) return;
+
+        requestAnimationFrame(() => {
```
kKaskak: You can use `onTransitionEnd` from the `Transition` component to avoid using workarounds. Invoke the scroll after it ends.

kKaskak (changes requested):

src/routes/Player/SideDrawer/SideDrawer.tsx
mesalilac: I couldn't find `onTransitionEnded` anywhere! 
could you provide more info

kKaskak: https://github.com/Stremio/stremio-web/blob/development/src/components/Transition/Transition.tsx#L48

Misspelled the function name sorry

mesalilac: when i try `onTransitionEnded` insted of `onTransitionEnd`
```tsx
        <div ref={ref} className={classNames(styles['side-drawer'], className)} onTransitionEnded={jumpToPlayingNow} onMouseDown={onMouseDown}>

```
Throws: "Warning: Unknown event handler property `onTransitionEnded`. It will be ignored."

mesalilac: > https://github.com/Stremio/stremio-web/blob/development/src/components/Transition/Transition.tsx#L48
> 
> Misspelled the function name sorry

Sorry i don't know what you mean
do i use the `ref` passed to the `SideDrawer` to do this in side a useEffect?
```tsx
        ref?.addEventListener('transitionend', onTransitionEnd);
        return () => ref?.removeEventListener('transitionend', onTransitionEnd);
```
could you show me a code example

kKaskak: > > https://github.com/Stremio/stremio-web/blob/development/src/components/Transition/Transition.tsx#L48
> > Misspelled the function name sorry
> 
> Sorry i don't know what you mean do i use the `ref` passed to the `SideDrawer` to do this in side a useEffect?
> 
> ```tsx
>         ref?.addEventListener('transitionend', onTransitionEnd);
>         return () => ref?.removeEventListener('transitionend', onTransitionEnd);
> ```
> 
> could you show me a code example

I was trying to point you in the correct direction so you can think about the solution yourself. The idea is to track when the transition on the `Transition` component ends and then pass it as a prop to the `SideDrawer` and invoke the scroll when its true (meaning the transition ended). You don't need to add `onTransitionEnded` to the `SideDrawer`. Try to think how you can track the transition timing and pass it to the `SideDrawer` as a `boolean` value to then do something like this: 

```
    useEffect(() => {
        if (transitionEnded) {
            selectedVideoRef.current?.scrollIntoView({ behavior: 'smooth', block: 'center' });
        }
    }, [transitionEnded]);
```

kKaskak (approved):

kKaskak:
fyi you can look up the code to see how much cleaner the solution i tried to explain to you is. 
1. Track when the transition ends with a state or variable inside `Transition` component.
2. Pass it to `SideDrawer` component
3. Invoke scroll.

That's it 🎊 

Please test it as well 👍🏼

mesalilac:
a3c895dfc6589e6b0083061885cbcd523e2ea181. Thank you, that was the obvious change that i should have made

mesalilac:
I have tested in "Movies", "Series", and "Channels". it worked perfectly with no errors.

kKaskak:
Great

PR #970, opened 2025-07-04
Player: Add subtitles size shortcuts

tymmesyde:


TRtomasz (approved):
on the deployed branch translations are missing

PR #974, opened 2025-07-15
Details: Scroll to last watched video

tymmesyde:


kKaskak (dismissed):

TRtomasz (dismissed):
can we add `on hover` effect on the element when scrolling to it? there is no indication which one is really the last one watched :D i see list scrolling but without and visual effect it is not possible to tell which one was the last one

kKaskak:
Good idea

CDrosos:
Beautiful feature, cant wait for it

CDrosos:
@kKaskak in the last edit, With the last check it will scroll only if the video is marked as watched? What if we are rewatching an episode. It should respect the scroll to last watched video, but i mean the video we were just last streaming, either is marked as watched or not or if we saw some minutes only otherwise?
Or is it a bad idea?

kKaskak:
> With the last check it will scroll only if the video is marked as watched? What if we are rewatching an episode. It should respect the scroll to last watched video, but i mean the video we were just last streaming, either is marked as watched or not or if we saw some minutes only otherwise? Or is it a bad idea?

the idea here is to prevent the scroll in cases where we watched all videos in a season
The check means that we will scroll when the video is **not** marked as watched.

CDrosos:
Yes yes sorry. Great edit then. 🙏

kKaskak (approved):

PR #979, opened 2025-07-30
Settings: Save playbackSpeed and volume

Aztup:
This PR depends on https://github.com/Stremio/stremio-core/pull/838


tymmesyde:
Hi, thank you for your PR,
why did you choose to pass `playbackSpeed` as arg to the load command of stremio-video?

Aztup:
Hello, the issue was that calling `video.load` resets the playback speed back to 1, and I believe there's no way to know when the video is actually loaded, but maybe there's a better way to do this?

elpiel:
Could this be added to the StreamItemState?
https://github.com/Stremio/stremio-core/blob/dcbb760af5f95b354b6ce6d8783fdb1c26bf34ec/src/types/streams/streams_item.rs#L29

This struct should keep track the settings per stream and bindge-watching group.

Aztup:
It's added. If I understand correctly, this is used to sync the state between devices? In that case, do I also need to add it somewhere else?

tymmesyde:
> Hello, the issue was that calling `video.load` resets the playback speed back to 1, and I believe there's no way to know when the video is actually loaded, but maybe there's a better way to do this?

Isn't it the same for `volume`, wouldn't it reset too?
You can check `video.state.loaded === true` to know if the video is loaded

Aztup:
No, volume would not reset, only the playback speed. I'm also pretty sure I did try to do it after it was loaded but it didn't seem to work.

tymmesyde:
> No, volume would not reset, only the playback speed. I'm also pretty sure I did try to do it after it was loaded but it didn't seem to work.

Can you pass either both `volume` and `playbackSpeed`  in the command options or find another way with react hook to set those?

Aztup:
This no longer requires https://github.com/Stremio/stremio-video/pull/117 so I've closed it. This version should be cleaner.

kKaskak:
hi pls resolve conflicts

Aztup:
Should be good :)

kKaskak:
outdated, we will do this in the streams bucket instead

PR #981, opened 2025-08-05
Update SearchParamsHandler.js

asnaek:
Currently the streamingServerUrl query will add the new server URL, but under Settings->Streaming section the new URL is not shown, in the selection list with status indicator. As a result the yellow popup still appears stating that the server cannot be reached.

image


tymmesyde (approved):

tymmesyde:
Thank you for your contribution!

PR #990, opened 2025-08-25
MetaPreview: MetaActions Button positioning fix

kKaskak:
check `@phone-landscape` before and after the change styles + mobile icon fixes


Botsy (approved):
I tested and it looks good 👍

PR #991, opened 2025-08-25
Search: dynamic view calculation fix on PWA

kKaskak:
on search page the content height was not calculated correctly which caused the navbar to not follow the safe area insets on iOS PWA


Botsy (approved):
Looks good 👍

PR #998, opened 2025-09-10
Feat/stream converted source

elpiel:


src/routes/Player/Player.js
```diff
@@ -322,10 +322,11 @@ const Player = ({ urlParams, queryParams }) => {
         setError(null);
         video.unload();
 
-        if (player.selected && streamingServer.settings?.type !== 'Loading') {
+        // if (player.stream?.type === 'Ready' && streamingServer.settings?.type !== 'Loading') {
```
roger18gm: Any particular reason for leaving this comment here or can it be removed for clarity?

elpiel: yes, it will be removed once this approach has been confirmed for the feature.

kKaskak (approved):

kKaskak (approved):

PR #1002, opened 2025-09-17
Intro: make all text lowercase to match the rest

### LINKED ISSUE #1000: [Bug]: On new user/register page, some buttrons are uppercase, some not
### Stremio-Web Version

/development branch

### Browser

Chrome

### Platform / Device type

Windows

### What Happened?

Also this string "must accept terms..." is not appear translated while it seems to exist on el-GR file as MUST_ACCEPT_TERMS
https://snipboard.io/65TG3f.jpg

### Logs

```shell

```

### Notes

_No response_

### Code of Conduct

- [x] I agree

ASiD-0:
## What I changed
- Converted button text to lowercase where needed in Intro.js to ensure consistent casing on the page

## Why
- Some buttons had uppercase text and some had title text

## How I tested
- Manually verified the sign up/login page in a dev build.
- Confirmed buttons now display with consistent casing

## Files changed
- 'Intro.js'

## Related issue
- **Closes #1000**

### DIFF
--- a/src/routes/Intro/Intro.js
+++ b/src/routes/Intro/Intro.js
@@ -387,7 +387,7 @@ const Intro = ({ queryParams }) => {
                     {
                         state.form === SIGNUP_FORM ?
                             <Button className={classnames(styles['form-button'], styles['login-form-button'])} onClick={switchFormOnClick}>
-                                <div className={classnames(styles['label'], styles['lowercase'])}>{t('LOG_IN')}</div>
+                                <div className={styles['label']}>{t('LOG_IN')}</div>
                             </Button>
                             :
                             null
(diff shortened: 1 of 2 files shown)


kKaskak:
here the only thing you need to do i just remove the class entirely from `Intro.js` and `styles.less` which is corresponding to `Intro.js`

src/routes/Intro/Intro.js
```diff
@@ -387,23 +387,23 @@ const Intro = ({ queryParams }) => {
                     {
                         state.form === SIGNUP_FORM ?
                             <Button className={classnames(styles['form-button'], styles['login-form-button'])} onClick={switchFormOnClick}>
-                                <div className={classnames(styles['label'], styles['uppercase'])}>{t('LOG_IN')}</div>
+                                <div className={classnames(styles['label'], styles['lowercase'])}>{t('LOG_IN')}</div>
```
kKaskak: here the only thing you need to do i just remove the class entirely from `Intro.js` and `styles.less` which is corresponding to `Intro.js`
```suggestion
                                <div className={styles['label']}>{t('LOG_IN')}</div>
```

kKaskak (changes requested):
and same applies to all cases

src/routes/Intro/Intro.js
ASiD-0: Okk I will do that.

kKaskak: please also remove unused class from `styles.less` as i mentioned before

ASiD-0: Apologies, should be good now

kKaskak (approved):

PR #1015, opened 2025-10-06
fix: include core fix for lib. item CW !removed check

elpiel:
- https://github.com/Stremio/stremio-core/pull/860


tymmesyde:
Core was updated on development including this change

PR #1016, opened 2025-10-06
Dev: Improve error logs

ckorber:
As mentioned in #296 error cast are now added with cause argument.

### DIFF
--- a/src/services/Shell/Shell.js
+++ b/src/services/Shell/Shell.js
@@ -63,7 +63,7 @@ function Shell() {
         } catch (e) {
             console.error(e);
             active = false;
-            error = new Error(e);
+            error = new Error('Failed to initialize shell transport', { cause: e });
             starting = false;
             onStateChanged();
             transport = null;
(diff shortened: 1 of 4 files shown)


src/services/Shell/Shell.js
```diff
@@ -63,7 +63,7 @@ function Shell() {
         } catch (e) {
             console.error(e);
             active = false;
-            error = new Error(e);
+            error = new Error(e, { cause: e });
```
tymmesyde: The cause should not be the same as the error, here the error message should be something like: 'Failed to initialize shell transport'

kKaskak:
please resolve build issues

ckorber:
I resolved the build issues.

kKaskak (dismissed):

src/services/Shell/Shell.js
```diff
@@ -63,7 +63,7 @@ function Shell() {
         } catch (e) {
             console.error(e);
             active = false;
-            error = new Error(e);
+            error = new Error(e, { cause: 'Failed to initialize shell transport' });
```
tymmesyde: Those should be swapped https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#rethrowing_an_error_with_a_cause
```suggestion
            error = new Error('Failed to initialize shell transport', { cause: e });
```

ckorber: Thank you for the suggestion!

kKaskak (approved):

tymmesyde:
Thank you!

PR #1018, opened 2025-10-07
Dev: add option to disable service worker

### LINKED ISSUE #986: [Feature]: Ability to disable service worker
### Feature Description

I am using authentik as a middleware for stremio, but because the service worker caches /, the cached page loads and this prevents middlewares from loading.

When I first visit the site when the service worker isn't registered, the middleware works as expected and authentik prompts me to login. Once I'm logged in, I can only use the site for until that session lasts as once it expires, I cannot login since / is cached.

This is especially problematic when using `tsaridas/stremio-docker` since the server is on the same domain, and so requests to the server are broken since they expect an authenticated user. I had first created an issue there, and was directed here - for more context: https://github.com/tsaridas/stremio-docker/issues/110

### Proposed Solution

An environment variable or something similar to disable the service worker, if possible. At least for root page?

_No response_

### Code of Conduct

- [x] I agree

NachoLZ:
### The Problem

When using middleware for authentication (e.g., authentik), the service worker's caching of the root page (`/`) can interfere with the authentication flow. If a user's session expires, the cached page is served, preventing the middleware from redirecting to the login page. This is particularly problematic in environments where the app and server are on the same domain.

This change addresses the issue reported in [GitHub Issue #986](https://github.com/Stremio/stremio-web/issues/986).

### The Solution

A `SERVICE_WORKER_DISABLED` environment variable has been added. When this variable is set to `'true'`, the application will skip the registration of the service worker. This allows authentication middleware to function as expected without interference from the cached root page.

### How to Use

To disable the service worker, set the following environment variable in your build environment:

`SERVICE_WORKER_DISABLED=true`

By default, if the variable is not set or is set to `false`, the service worker will be registered as usual.

### Code of Conduct
- [x] I agree

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -36,7 +36,7 @@ i18n
 const root = ReactDOM.createRoot(document.getElementById('app'));
 root.render(<App />);
 
-if (process.env.NODE_ENV === 'production' && 'serviceWorker' in navigator) {
+if (process.env.NODE_ENV === 'production' && 'serviceWorker' in navigator && process.env.SERVICE_WORKER_DISABLED !== 'true') {
     window.addEventListener('load', () => {
         navigator.serviceWorker.register('service-worker.js')
             .catch((registrationError) => {
--- a/src/index.js
+++ b/src/index.js
@@ -36,7 +36,7 @@ i18n
 const root = ReactDOM.createRoot(document.getElementById('app'));
 root.render(<App />);
 
-if (process.env.NODE_ENV === 'production' && 'serviceWorker' in navigator && process.env.SERVICE_WORKER_DISABLED !== 'true') {
+if (process.env.NODE_ENV === 'production' && process.env.SERVICE_WORKER_DISABLED !== 'true' && process.env.SERVICE_WORKER_DISABLED !== true && 'serviceWorker' in navigator) {
     window.addEventListener('load', () => {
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -36,7 +36,7 @@ i18n
 const root = ReactDOM.createRoot(document.getElementById('app'));
 root.render(<App />);
 
-if (process.env.NODE_ENV === 'production' && 'serviceWorker' in navigator) {
+if (process.env.NODE_ENV === 'production' && 'serviceWorker' in navigator && process.env.SERVICE_WORKER_DISABLED !== 'true') {
```
tymmesyde: in this case if you set `SERVICE_WORKER_DISABLED` to `true` in webpack.config.js, it will not disable it because it's parsed as boolean instead of a string
```suggestion
if (process.env.NODE_ENV === 'production' && process.env.SERVICE_WORKER_DISABLED !== 'true' && process.env.SERVICE_WORKER_DISABLED !== true && 'serviceWorker' in navigator) {
```

NachoLZ: Great catch, thank you! I
I've committed your suggestion.

tymmesyde (approved):

tymmesyde:
Thank you!

xam1dullo (commented):
ads

PR #1025, opened 2025-10-08
Player: Add support for media play/pause keys

PsyGuy007-sys:
## Summary
- handle MediaPlayPause, MediaPlay and MediaPause codes
- reuse existing play/pause toggle logic from the player

## Testing
- not covered

### DIFF
--- a/src/routes/Player/Player.js
+++ b/src/routes/Player/Player.js
@@ -547,6 +547,36 @@ const Player = ({ urlParams, queryParams }) => {
 
                     break;
                 }
+                case 'MediaPlayPause': {
+                    if (!menusOpen && !nextVideoPopupOpen && video.state.paused !== null) {
+                        event.preventDefault();
+                        if (video.state.paused) {
+                            onPlayRequested();
+                            setSeeking(false);
+                        } else {
+                            onPauseRequested();
+                        }
+                    }
+
+                    break;
+                }
+                case 'MediaPlay': {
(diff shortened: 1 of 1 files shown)


PsyGuy007-sys:
Thanks for taking a look! Ready for review.

kKaskak (approved):
code seems ok but i do not have a keyboard with these to test maybe @tymmesyde 

---


will try testing with airpods

---

ok tested with sony headphones and it seems to work fine


---
but it seems to work even without this change on my headphones :D

PsyGuy007-sys:
I tried it myself and yeah, it actually works even without the change... I should’ve checked before pushing that bit of code lol 

> code seems ok but i do not have a keyboard with these to test maybe @tymmesyde
> 
> will try testing with airpods
> 
> ok tested with sony headphones and it seems to work fine
> 
> but it seems to work even without this change on my headphones :D

kKaskak:
> I tried it myself and yeah, it actually works even without the change... I should’ve checked before pushing that bit of code lol
> 
> > code seems ok but i do not have a keyboard with these to test maybe @tymmesyde
> > will try testing with airpods
> > ok tested with sony headphones and it seems to work fine
> > but it seems to work even without this change on my headphones :D

it's okay these are still valid cases for older devices so i don't see a reason why to not add these

Ejakod:
It still does not work using Stremio 5 on MacOs (silicon)

Happy to test with a test build if needed

PR #1026, opened 2025-10-08
Misc: Use correct args for PullUserFromAPI method

kKaskak:
because of recent core changes we can't skip `args` even when empty


kKaskak:
duplicate 🗡️

Issue #1019, opened 2025-10-07
[Feature]: Adding more documentation


onel:
### Feature Description

I would love to contribute and add more technical docs to the project. 
Is the project open to receiving some PRs in that sense?

### Code of Conduct

- [x] I agree

kKaskak:
What kind of documentation are you talking about?

Rakeshvarma007:
he is talking about the documentation for understanding of the code

onel:
Yup exactly, mainly developer documentation.
Let me start with adding some reference docs.

Any areas or specific types of docs that are the most needed?

Nayab-Gauhar:
Hi! I'd like to work on this issue for Hacktoberfest. Could you please assign it to me?

onel:
@Nayab-Gauhar you can work on this without being assigned. Just create a PR and mention this issue in it.

PR #1048, opened 2025-10-16
Added reference docs

onel:
This pull request adds new reference documentation for a couple of files.
The branch was created from `development`.

Added reference documentation to the following files:

- src/routes/Player/SideDrawer/SideDrawer.tsx
- src/common/Tooltips/Tooltip/Tooltip.js
- src/routes/MetaDetails/VideosList/VideosList.js

#1019

### DIFF
--- a/src/common/Tooltips/Tooltip/Tooltip.js
+++ b/src/common/Tooltips/Tooltip/Tooltip.js
@@ -5,32 +5,56 @@ const PropTypes = require('prop-types');
 const useTooltip = require('../useTooltip');
 const styles = require('./styles');
 
+/**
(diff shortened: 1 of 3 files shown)


onel:
The CLA is signed but the @CLAassistant might still flag this as not because the commit author is not me personally.
I've encountered this problem with other repos as well.

This is not an issue with me with you guys are also ok with it.

kKaskak:
there is absolutely no reason for docs like these in UI files. If you are a developer you know what a function does right away. 
The function names are self explanatory.

onel:
Ok, that makes sense. What would be some good parts of the repo that really need reference docs? @kKaskak

PR #1059, opened 2025-10-30
Details: improve selected video logic

kKaskak:


src/components/Video/styles.less
```diff
@@ -172,6 +176,20 @@
...
+    @keyframes border {
+        0% {
+            border: 0.15rem solid var(--primary-accent-color);
+        }
+        100% {
+            border: 0.15rem solid transparent;
+        }
+    }
+
+
```
Botsy, line 192: Shouldn't we leave the border visible so that the selected episode stays highlighted? Otherwise, I think this animation might be missed. Maybe we can keep the animation, just instead of making it entirely transparent in the end, we could go with lower opacity so that the border is not that bright, but it's still there, as an accent to the selected video. What do you think?

kKaskak: @alesinka any thoughts? you can check the PR here
https://stremio.github.io/stremio-web/feat/details-selected-video-styles/

TRtomasz (approved):

Issue #1068, opened 2025-11-22
[Feature]: Touch & hold to play video at 2x speed


MixologyXBot:
### Feature Description

**Summary**

It would be very helpful if the video player supported a “touch and hold to speed up” interaction.  
When the user touches and holds on the video, playback should temporarily switch to 2x speed, and when the user releases, it should return to the original speed.

**Why this is useful**

- Quickly skim through less important parts of a video without changing the global playback speed setting.  
- More convenient on mobile, where opening a speed menu or controls can be slower and less precise.  
- Matches the “press and hold to fast-forward” style interaction that many users are already familiar with.

**Proposed behavior**

- On mobile (and possibly desktop with mouse button held down):
  - When the user **touches and holds** on the video area (or a specific control), the playback speed becomes **2x**.
  - When the user **releases**, the playback speed goes back to the previous value (e.g., 1x or whatever the user had set).
- This should not permanently change the saved playback speed setting.

**Optional enhancements**

- Allow customizing the temporary speed (e.g., 1.5x, 2x, 3x) via settings.
- Visual indication (e.g., a small “2x” badge) while the user is holding the screen.

**Example user flow**

1. User watches the video at normal speed (1x).
2. User wants to quickly pass through a slow section.
3. User presses and holds on the video → playback becomes 2x while holding.
4. User releases → playback returns to 1x.

Thank you for considering this feature!

### Code of Conduct

- [x] I agree

Haidarabbas731:
can you please assign this issue to me?

ramsyjonas:
this feature would be great

PR #1088, opened 2025-12-11
feature: speeds up video when holding MLB

Pranav1703:
[#1068](https://github.com/Stremio/stremio-web/issues/1068)
Adds a feature where user clicks and holds mouse left button on video player speeds up playback to 2x and when released, the playback speed is reverted to its previous value.<br>
Couldn't test this feature on mobile because i keep getting 'Video is not supported' msg.
its working for pc tho:)

### DIFF
--- a/src/routes/Player/Player.js
+++ b/src/routes/Player/Player.js
@@ -47,6 +47,10 @@ const Player = ({ urlParams, queryParams }) => {
     const toast = useToast();
 
     const [seeking, setSeeking] = React.useState(false);
+    
+    const isMlbHeld = React.useRef(false);
+    const MlbHoldTimerRef = React.useRef(null);
+    const currPlaybackSpeed = React.useRef(null)
 
     const [casting, setCasting] = React.useState(() => {
         return services.chromecast.active && services.chromecast.transport.getCastState() === cast.framework.CastState.CONNECTED;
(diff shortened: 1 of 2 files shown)


kKaskak (commented):
## Review

This feature has already been implemented on the `development` branch (mouse hold + spacebar hold for 2x speed), and PR #1173 is currently open fixing bugs in that implementation. This PR is superseded and can be closed.

For reference, the differences from what landed on `development`:
- `development` uses `window` event listeners instead of React event props on `<Video>`, which also handles spacebar hold
- `development` uses 200ms delay vs 500ms here
- `development` routes through `onPlaybackSpeedChanged` instead of calling `video.setProp` directly
- `development` doesn't require modifying `Video.js` props

### Issues that would apply if this weren't superseded

- **`video.setProp('playbackSpeed', ...)` called directly** instead of going through `onPlaybackSpeedChanged` — bypasses any side effects or state sync that callback handles
- **`Video.js` modified to pass individual mouse event props** — fragile, would need updating for each new event. The `development` approach of using `window` listeners is cleaner
- **No spacebar hold support** — only mouse, `development` has both
- **Naming**: `MlbHoldTimerRef` uses PascalCase for a ref (should be camelCase), `currPlaybackSpeed` abbreviates unnecessarily

PR #1097, opened 2025-12-18
Dev: use shortcuts provider on player

tymmesyde:


kKaskak:
commit `c9a40aabd7db97fb71b99f5fb9887f413c0901fc` broke statistics menu handler

While porting the D-key (statisticsMenu) handler, the condition
  check on the stream was garbled. Before:                                                                                                                          
                                                            
 ```
 player.selected && typeof player.selected.stream.infoHash === 'string' && typeof player.selected.stream.fileIdx === 'number'       
 ```                               
                                              
  After — stream got extracted into a local and the .infoHash / .fileIdx accessors were dropped, leaving a self-contradictory check:                                
```
  const stream = player.selected?.stream;                                                                                                                           
  ... && typeof stream === 'string' && typeof stream === 'number'       
  ```                                                                                            
                                                                                                                                                                    
D shortcut has been dead from Dec 15 → present.

PR #1101, opened 2025-12-25
Settings: Add discord rich presence

kKaskak:


src/routes/Player/Player.js
```diff
@@ -540,6 +542,39 @@ const Player = ({ urlParams, queryParams }) => {
...
+        if (video.state.stream === null || typeof player?.title !== 'string') {
+            discordStartTimestamp.current = null;
+            discord.setActivity(null);
+            return;
+        }
+
+        if (video.state.paused) {
+            discordStartTimestamp.current = null;
+        } else if (typeof video.state.time === 'number') {
+            const startTimestamp = Math.floor((Date.now() / 1000) - video.state.time);
```
Botsy: This calculation is not correct, `video.state.time` is in milliseconds so please fix it to `Math.floor((Date.now() - video.state.time) / 1000);`

Botsy: Otherwise it shows "Watching 67:14:00" when the video has been playing for ~4 minutes.

src/services/index.js
```diff
@@ -1,11 +1,13 @@
 // Copyright (C) 2017-2023 Smart code 203358507
 
 const Chromecast = require('./Chromecast');
+const Discord = require('./Discord');
```
Botsy: please fix the import to use the default import - `const { default: Discord } = ..`, right now it breaks the app

src/services/Discord/Discord.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    clearActivity(): void {
+        if (this.shell && this.shell.active) {
+            this.shell.send('discord-clear-activity', {});
+        }
+    }
+
+    get available(): boolean {
+        return this.shell && this.shell.active;
+    }
```
Botsy, line 71: What happens if user doesn't have Discord installed at all? Shouldn't the `available` flag reflect whether Discord is actually usable (after a successful connection)? And Settings toggle should show and display connection state more clearly (connected / disconnected / unavailable). Probably we need some feedback when connection is successful/failed in toast message as well?

Botsy (changes requested):
I tested with suggested fixes and it worked for me. Still I think we might need some improvements in terms of giving feedback to the user when a connection is successful or has failed/is unavailable.

Botsy (approved):
lgtm

tymmesyde:
Wanted to comment but this got merged before i could, having both a service and a provider doesn't make sense
the logic from the service should be in the provider, we don't need services anymore

kKaskak:
> Wanted to comment but this got merged before i could, having both a service and a provider doesn't make sense the logic from the service should be in the provider, we don't need services anymore

yup this is my mistake im fixing this now, pr is multiple months old and i did not recheck

PR #1105, opened 2025-12-28
feat: cross-season search for series

filipbuchta:
Search now works across all seasons instead of just current season 
Added season separators with highlighting for current season (only when search results are from different season)
Auto-scroll to selected season in search results (to keep the original behavior, you always see the results from selected season this way)

image

### DIFF
--- a/src/routes/MetaDetails/VideosList/VideosList.js
+++ b/src/routes/MetaDetails/VideosList/VideosList.js
@@ -72,6 +72,27 @@ const VideosList = ({ className, metaItem, libraryItem, season, seasonOnSelect, 
     }, [videosForSeason]);
 
     const [search, setSearch] = React.useState('');
+
+    const filteredVideos = React.useMemo(() => {
+        // When searching: use ALL videos across all seasons
+        // When browsing normally: use only videos from selected season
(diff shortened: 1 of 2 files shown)


kKaskak:
feature wise it looks nice, code wise we cannot merge it for now :D think of how we could improve code quality and the UX of the feature for users. for example switching seasons wont do anything if you have something in the search bar typed in already. Please check all edge cases that could happen because of this new improved feature

filipbuchta:
Search is now cleared when changing season

When there are no results in the selected season it will now still display the separator, so that you are aware that it's result from different season
image

tymmesyde:
Hi, please stop using LLMs to code, look at the code already there and try to match the pattern, currently this is very hard to read, there are many unnecessary comments, some logic is in the JSX while it should be outside of it

kKaskak:
> Hi, please stop using LLMs to code, look at the code already there and try to match the pattern, currently this is very hard to read, there are many unnecessary comments, some logic is in the JSX while it should be outside of it

Plus one on this

filipbuchta:
Sorry, I currently don't have time capacity to rewrite this without llm. I'll come back later!

PR #1107, opened 2026-01-10
feat: add multi-account switching support

maximedeprince:
## Description
This PR adds multi-account switching support to Stremio Web, allowing users to store and switch between multiple Stremio accounts locally without having to log out and re-enter credentials each time.

---

### 👤 Account Management
* **AccountManager Utility**: A new utility that persists multiple accounts in `localStorage`.
* **Data Persistence**: Stores email, avatar, `authKey`, user settings, and last active timestamp per account.
* **Per-Account Settings**: Preferences (player, subtitles, audio language, etc.) are saved individually and restored automatically upon switching.

### 🖼️ Profile Picker UI
* **ProfilePicker Component**: Displayed on the intro page when multiple accounts are detected.
* **Visual List**: Shows all saved accounts with avatars and email-based display names.
* **Management Mode**: A "Manage Profiles" toggle allows for the deletion of saved accounts.
* **Add Profile**: Dedicated button to authenticate and link additional accounts.

### 🛠️ NavMenu Integration
* **Change Profile Link**: Added next to "Log out" in the user menu.
* **Smart Redirect**: "Change Profile" saves current settings before redirecting to the picker.
* **Smart Logout**: "Log out" removes only the current account and redirects to the picker if other profiles remain.

### 🛡️ Token & Security
* **Expiration Handling**: Gracefully handles expired or invalid auth tokens.
* **Auto-Cleanup**: Automatically removes invalid accounts from local storage.
* **Re-auth Helper**: Pre-fills the email field for convenient re-authentication.

---

## Technical Notes
* **Auth Core**: Uses the existing `LoginWithToken` auth type from `stremio-core` (no backend changes required).
* **Action Flow**: Settings are restored via the `UpdateSettings` action after a successful token login.
* **Routing**: Route protection has been updated to allow the `?profiles=true` parameter for authenticated users to access the profile picker.

---

## Screenshots
| Profile Picker | Nav Menu |
| :--- | :--- |
|image|image|

### DIFF
--- a/src/App/withProtectedRoutes.js
+++ b/src/App/withProtectedRoutes.js
@@ -15,7 +15,9 @@ const withProtectedRoutes = (Component) => {
             previousAuthRef.current = profile.auth;
         }, [profile]);
         const onRouteChange = React.useCallback((routeConfig) => {
-            if (profile.auth !== null && routeConfig.component === Intro) {
+            // Allow access to /intro when ?profiles=true (for profile switching)
+            const urlParams = new URLSearchParams(window.location.hash.split('?')[1]);
+            if (profile.auth !== null && routeConfig.component === Intro && urlParams.get('profiles') !== 'true') {
                 window.location.replace('#/');
                 return true;
             }
--- a/src/common/AccountManager.js
+++ b/src/common/AccountManager.js
@@ -1,0 +1,86 @@
+// Copyright (C) 2017-2026 Smart code 203358507
+
+const STORAGE_KEY = 'stremio-multi-accounts';
+
+const AccountManager = {
+    getAccounts: () => {
+        try {
+            return JSON.parse(localStorage.getItem(STORAGE_KEY)) || [];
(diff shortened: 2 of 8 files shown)


tymmesyde:
Hi, just want to point that the storage is managed by [stremio-core](https://github.com/Stremio/stremio-core) and not by the app, so `AccountManager.js` cannot be a thing
Also this feature requires much more planning with the team so it can't be implemented now

maximedeprince:
> Hi, just want to point that the storage is managed by [stremio-core](https://github.com/Stremio/stremio-core) and not by the app, so `AccountManager.js` cannot be a thing Also this feature requires much more planning with the team so it can't be implemented now

Ah, my bad! I didn't realize storage was strictly handled by the core side.

That makes total sense regarding cross-platform support. I'll close this for now then. Hopefully we can see this feature land properly in the future!

Thanks for the review

PR #1115, opened 2026-01-21
Player: Remember selected tracks

tymmesyde:


kKaskak:
In commit `39741d137` (feat: remember selected tracks on player), the auto-select effect was changed to prefer `streamState.audioTrack.id`. But if the saved ID no longer matches any current audio track (different stream source for the same episode, different transcoding, etc.), the code returns undefined
  and the settings.audioLanguage fallback is never tried — the player's default audio track plays instead.

PR #1120, opened 2026-01-28
Fix/Prevent autoplay when using an external player

ArtificialSloth:
The play in external device option pauses the web player when selected, but if the video is still loading then it can't be paused and it will just start playing like normal as soon as its done loading.

This adds a flag that's used to pause the web player if it starts playing after the user has opened the stream in an external player.

### DIFF
--- a/src/routes/Player/Player.js
+++ b/src/routes/Player/Player.js
@@ -92,6 +92,7 @@ const Player = ({ urlParams, queryParams }) => {
     const defaultSubtitlesSelected = React.useRef(false);
     const subtitlesEnabled = React.useRef(true);
     const defaultAudioTrackSelected = React.useRef(false);
+    const playingOnExternalDevice = React.useRef(false);
     const [error, setError] = React.useState(null);
 
     const isNavigating = React.useRef(false);
@@ -189,6 +190,7 @@ const Player = ({ urlParams, queryParams }) => {
     }, []);
 
     const onPlayRequested = React.useCallback(() => {
+        playingOnExternalDevice.current = false;
         video.setPaused(false);
         setSeeking(false);
     }, []);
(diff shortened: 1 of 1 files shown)


Botsy:
@ArtificialSloth could you describe the steps to reproduce the issue you faced before those changes? What external player are you using?

I couldn't reproduce the same issue with or without your changes, but I found another one when using an external player. On my side if I've already started watching a stream in web, even if I change the settings to use external player, the video won't start in the external player. But I assume this issue is unrelated.

ArtificialSloth:
Yah, I believe the other issue you found is unrelated. The one I'm trying to resolve is something I've experienced using Stremio in Firefox and Edge, and I've seen this issue when trying to play with both VLC and MPV. Although I don't think this issue is dependent on any specific external player.

To reproduce the issue:
Open any stream on web.stremio.com
Click Play in VLC/MPV immediately, while the web player is still loading the source (before [this](https://github.com/Stremio/stremio-video/blob/e52b3a690704374d35187d536ce3b2a7fe8828a9/src/HTMLVideo/HTMLVideo.js#L546C1-L546C43) promise resolves)
The video should open without issue in the external player, but as soon as the web player is done loading, the video starts playing.

My understanding of whats happening is a sort of race condition. video.load runs instantly with autoplay: true and a promise is created to fetch and set the source, then the PlayingOnDevice event fires before that promise resolves, calling videoElement.pause() on a video element that isn't playing and doesn't have a source so there's no effect, finally once the promise resolves and the source is set it respects autoplay: true and begins playing.

kKaskak:
i recon race condition im okay with this getting merged, however player refactoring will most likely eliminate this tho its not coming anytime soon :(

kKaskak:
i dont see a reason why not to merge this

kKaskak (commented):
## Review

Clean fix for the race condition where `PlayingOnDevice` fires `onPauseRequested()` but the video hasn't loaded yet, so the pause is a no-op and it auto-plays once ready.

### Looks good

- Flag set on `PlayingOnDevice`, checked when `video.state.paused` changes — catches the exact race ✓
- Flag cleared on `onPlayRequested` — user explicitly hitting play after casting works normally ✓
- Flag cleared on stream change — no stale state across navigations ✓

LGTM.

kKaskak (approved):

PR #1129, opened 2026-02-08
[Feature]: Add copy magnet link option

### LINKED ISSUE #1128: [Feature]: Copy magnet link from stream context menu
### Feature Description

When seeing the various streaming options for one particular movie/series, we can right click it and get these options:
 - **Play**
 - **Copy Stream Link**
 - **Copy Download Video Link**

But, there is currently no way to copy the actual magnet URI for a torrent stream. Having a "**Copy Magnet Link**" option would allow users to easily share the torrent or to open it in some torrent client, without having to manually look for the link.

This is especially useful, given that the download link, even though it allows to download the file, it forces the user to use the browser to mange its downloads, which is not ideal when comparing to a torrent client.

This applies to both the streaming options menu, as well as in the "..." options inside the actual player.

### Proposed Solution

Add a  "**Copy magnet link**" option that builds a magnet URI from the `infoHash` and `sources` fields already present in the stream data. 

The `magnet-uri` package, which is already a project dependency, can be used to encode the link. The result would look like:

```
magnet:?xt=urn:btih:<infoHash>&dn=<stream name>&tr=<tracker1>&tr=<tracker2>
```

The option would only appear for torrent-based streams (i.e. streams that have an `infoHash`). The existing `magnet-link` icon from `@stremio/stremio-icons` can be used for the menu entry.

in the stream options menu:

Image

inside the player:

Image

### Code of Conduct

- [x] I agree

duartebranco:
Closes #1128

Adds a "**Copy magnet link**" option to the stream context menu and to the player options menu.

Builds the magnet URI from the `infoHash` and `sources` fields already available in the stream data, using the existing `magnet-uri` dependency. The option only shows for torrent-based streams.

Uses the `magnet-link` icon from `@stremio/stremio-icons`.

Translation keys (`CTX_COPY_MAGNET_LINK`, `PLAYER_COPY_MAGNET_LINK_SUCCESS`, `PLAYER_COPY_MAGNET_LINK_ERROR`) use inline fallbacks until they are added to `stremio-translations`.

### DIFF
--- a/src/routes/MetaDetails/StreamsList/Stream/Stream.js
+++ b/src/routes/MetaDetails/StreamsList/Stream/Stream.js
@@ -3,6 +3,7 @@
 const React = require('react');
 const PropTypes = require('prop-types');
 const classnames = require('classnames');
+const magnet = require('magnet-uri');
 const { default: Icon } = require('@stremio/stremio-icons/react');
 const { t } = require('i18next');
 const { useProfile, usePlatform, useToast, useBinaryState } = require('stremio/common');
@@ -12,7 +13,7 @@ const { useRouteFocused } = require('stremio-router');
 const StreamPlaceholder = require('./StreamPlaceholder');
 const styles = require('./styles');
 
-const Stream = ({ className, videoId, videoReleased, addonName, name, description, thumbnail, progress, deepLinks, ...props }) => {
+const Stream = ({ className, videoId, videoReleased, addonName, name, description, thumbnail, progress, deepLinks, infoHash, sources, ...props }) => {
     const profile = useProfile();
     const toast = useToast();
     const platform = usePlatform();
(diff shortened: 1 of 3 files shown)


duartebranco:
Could someone clarify why this PR was closed? 

I’d like to understand what changes would be needed to align it with the project’s direction.

kKaskak:
> Could someone clarify why this PR was closed?
> 
> I’d like to understand what changes would be needed to align it with the project’s direction.

its because we already support the magnet link in core and you should have used it, its already implemented fyi there is no need to use any packages nor do any of the logic you did in the UI code

kKaskak:
code looks ai generated (inline func resolving, inline translations instead of pkg update with pr, no understanding how core works)

kKaskak:
thank you for the PR, you can make another correct one if you would like, however this needs to be merged 
https://github.com/Stremio/stremio-core/pull/906

PR #1133, opened 2026-02-14
feat(Discover): show loading bar during pagination

klnstprx:
## Summary
- Render an animated loading bar at the bottom of the Discover catalog grid while the next page is being fetched
- Driven by the new `loadingNextPage` field from stremio-core's Discover serializer
- Includes `prefers-reduced-motion` fallback

Depends on: Stremio/stremio-core#932 (core-web release with `loadingNextPage`)


kKaskak:
i guess cool but it's not required and moves content and cells to the top, this was never requested by anyone, nor needed since most of the time the content loads before you even scroll to the bottom of the already loaded one, code wise looks ok but designer team rejected this thank you for the PR

PR #1140, opened 2026-02-23
PWA: navigation bar styles, fullscreen, safe areas

kKaskak:
- after ios26 the bar went back to white
- the safe area inset was white causing the app to look bad :D
- SideDrawer did not respect safe areas


Botsy (dismissed):
tested and all looks good 👍

Botsy (dismissed):

PR #1141, opened 2026-02-23
Dev: fix webpack config upgrade cross origin requests

kKaskak:
potential issues on iOS PWA app where addons are blocked from receiving the request


kKaskak:
non reproducible

PR #1151, opened 2026-03-04
MetaDetails: Add Copy Magnet Link option

### LINKED ISSUE #1128: [Feature]: Copy magnet link from stream context menu

kKaskak:


Botsy (changes requested):
Looks good. What about missing translations?

kKaskak:
@Botsy review now? thx

TRtomasz (dismissed):

Botsy (dismissed):

Issue #1132, opened 2026-02-14
Dockerfile doesnt build - slash missing in CP command


dominikvpb:
Buildfile is currently not working. The CP command has to have the target destination with a / at the end.

`COPY package.json pnpm-lock.yaml /var/www/stremio-web/`

https://github.com/Stremio/stremio-web/blob/a77faea0b9e6f06ca49777ecd168a1f50b88ff6e/Dockerfile#L22

kKaskak:
> Buildfile is currently not working. The CP command has to have the target destination with a / at the end.
> 
> `COPY package.json pnpm-lock.yaml /var/www/stremio-web/`
> 
> [stremio-web/Dockerfile](https://github.com/Stremio/stremio-web/blob/a77faea0b9e6f06ca49777ecd168a1f50b88ff6e/Dockerfile#L22)
> 
> Line 22 in [a77faea](/Stremio/stremio-web/commit/a77faea0b9e6f06ca49777ecd168a1f50b88ff6e)
> 
>  COPY package.json pnpm-lock.yaml /var/www/stremio-web


```
=> [app 3/4] COPY . /var/www/stremio-web                                                                                                                           8.9s 
 => [app 4/4] RUN pnpm build                                                                                                                                       15.8s 
 => [stage-3 3/3] COPY --from=app /var/www/stremio-web/build /var/www/stremio-web/build                                                                             0.0s 
 => exporting to image                                                                                                                                              0.1s 
 => => exporting layers                                                                                                                                             0.1s 
 => => writing image sha256:                                                                       0.0s 
                                                                                                                                                                         
View build details: docker-desktop://dashboard/build/desktop-linux/desktop-linux/  
```

kKaskak:
if im missing smth pls feel free to correct me, i tested and ran locally everything worked

dominikvpb:
Sorry, my bad. It seems to be a problem on that specific server somehow. Build script exited for me with error on
`[app 1/4] COPY package.json pnpm-lock.yaml /var/www/stremio-web` if no slash at the end.

On a fresh Ubuntu 24.4 VPS everything is working fine. Cheers!

PR #1157, opened 2026-03-12
Fix Dockerfile COPY command path for stremio-web

cnrbzk:


kKaskak:
explanation 
https://github.com/Stremio/stremio-web/issues/1132

PR #1163, opened 2026-03-18
feature - Add addons reorder capabilities

matthieume:
Add addon reorder capabilities in the web version.
Only available when logged in.

Requires an update in the core beforehand : https://github.com/Stremio/stremio-core/pull/946 

There is currently no way to reorder addons within Stremio. Adding this capability to the web version would be a significant improvement.
Today, users rely on external plugins, websites, or tools to reorder their addons. This approach is inherently risky, as these third‑party solutions cannot be fully trusted.

image

### DIFF
--- a/src/routes/Addons/Addon/Addon.js
+++ b/src/routes/Addons/Addon/Addon.js
@@ -8,7 +8,7 @@ const { default: Icon } = require('@stremio/stremio-icons/react');
 const { Button, Image } = require('stremio/components');
 const styles = require('./styles');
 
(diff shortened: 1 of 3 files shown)


kKaskak:
Thank you for your submission. While we appreciate the thought you put into this PR sadly this won't work correctly because of different reasons. Rearranged addons can break certain app's functionality entirely. There are a lot of users that modify the manifests or addons list which then leads to broken accounts. Out of nowhere their Stremio doesn't work anymore.

kKaskak:
FYI we are already working on a solution for this that will allow safe re-ordering of addons 🎉

PR #1172, opened 2026-03-25
fix: shift subtitles above control bar when controls are visible

Cleiton366:
## Problem
When moving the mouse, the control bar appears at the bottom of the player and overlaps the subtitles, making them unreadable. This is especially disruptive when watching foreign language content, since the bar has a delay before hiding and context can be lost.

## Solution
When the control bar becomes visible, subtitles are shifted upward to avoid the overlap. When the bar hides, subtitles return to their original position.

Additionally, subtitle elements were set to `pointer-events: none` to prevent them from interfering with the bar's auto-hide timer.

## Changes
- `Player.js`: added effect to adjust subtitle offset based on control bar visibility
- `Video/styles.less`: added `pointer-events: none` to video children

## Proof

Dev branch:
image
image

Feat branch:
image
image

### DIFF
--- a/src/routes/Player/Player.js
+++ b/src/routes/Player/Player.js
@@ -510,9 +510,13 @@ const Player = ({ urlParams, queryParams }) => {
     }, [video.state.stream, player.streamState]);
 
     React.useEffect(() => {
-        const base = settings.subtitlesOffset ?? 0;
-        video.setSubtitlesOffset(overlayHidden ? base : Math.min(100, base + CONTROL_BAR_SUBTITLE_BUMP));
-    }, [overlayHidden, settings.subtitlesOffset]);
+        if (video.state.stream === null || video.state.subtitlesOffset === null) return;
+        const base = player.streamState?.subtitleOffset ?? settings.subtitlesOffset ?? 0;
+        const desired = overlayHidden ? base : Math.min(100, base + CONTROL_BAR_SUBTITLE_BUMP);
+        if (video.state.subtitlesOffset !== desired) {
+            video.setSubtitlesOffset(desired);
+        }
+    }, [overlayHidden, settings.subtitlesOffset, player.streamState, video.state.subtitlesOffset, video.state.stream]);
 
     React.useEffect(() => {
         defaultSubtitlesSelected.current = false;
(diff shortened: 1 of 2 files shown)


kKaskak (commented):
## Review

Good idea — subtitles overlapping the control bar is a real UX pain point. The `pointer-events: none` fix is clean. However the subtitle offset approach has some issues.

### Issues

**1. Feedback loop / conflicting writes to `subtitlesOffset`**

The user can manually change subtitle offset via `onSubtitlesOffsetChanged` (line 284-287), which calls both `video.setSubtitlesOffset(offset)` and `streamStateChanged({ subtitleOffset: offset })` to persist it. Your effect also calls `video.setSubtitlesOffset()` but doesn't persist via `streamStateChanged` — so the user's chosen offset and the "bumped" offset fight each other.

Worse, `video.state.subtitlesOffset` is in your effect's dependency array. When the user changes the offset via the subtitles menu, your effect fires, sees the new value, and may immediately override it with `base + CONTROL_BAR_SUBTITLE_BUMP` if the overlay is visible. The user would never be able to adjust offset while controls are showing.

**2. `player.streamState?.subtitleOffset` vs `settings.subtitlesOffset` — which is "base"?**

```js
const base = player.streamState?.subtitleOffset ?? settings.subtitlesOffset ?? 0;
```

`streamState.subtitleOffset` is the per-stream override set when the user adjusts offset. `settings.subtitlesOffset` is the global default. But both of these already get applied to the video at lines 106 and 504-506. Your effect adds `CONTROL_BAR_SUBTITLE_BUMP` on top — but if the user then adjusts offset (which persists via `streamStateChanged`), the bumped value could leak into the persisted state on the next cycle.

**3. Hardcoded bump value**

`CONTROL_BAR_SUBTITLE_BUMP = 8` is a fixed pixel/percentage offset. The actual control bar height varies (it has a seek bar, buttons, etc.) and could change. A CSS-based solution would be more robust — shifting subtitles via a CSS class/variable on the video container when controls are visible, rather than mutating the video's subtitle offset property.

### Suggestion

Consider a CSS-based approach instead — add a class to the subtitle container (or a CSS custom property) that shifts subtitles up via `transform` or `bottom` when the overlay is visible. This avoids all the state conflicts above and doesn't touch the user's subtitle offset setting at all. Something like adding a `data-controls-visible` attribute or class to the video container and using CSS to shift subtitles.

### Looks good

- `pointer-events: none` on video children — clean fix to prevent subtitles from blocking the auto-hide timer ✓
- Guarding on `video.state.stream === null` and `subtitlesOffset === null` ✓

kKaskak:
for clarity i will close this PR, but if you ping me with the necessary changes i will re-review and open if everything checks out

PR #1173, opened 2026-03-26
Player: Fix spacebar and left click hold behavior

### LINKED ISSUE #2464: Pressing spacebar key resets playback speed
### Stremio Platform

macOS

### Have you tried play this video in external player

Yes

### Describe the bug

After setting a custom playback speed, when I press spacebar it automatically resets playback speed back to normal. Pressing spacebar for the second time then pauses the video.

### To Reproduce

1. Play a video
2. Change its playback speed
3. Press space to pause the video. The video won't pause but the speed will reset to 1x.
4. If you press space again, the video will pause

### Expected behavior

The playback speed should stay on the set value, not reset when pressing a pause shortcut 

### Screenshots

_No response_

### App Version

5.0.0-beta.32

### Device Info

Mac Mini M4 Pro

### OS Version

MacOS Tahoe 26.4

### Player Info

_No response_

### Browser Info

_No response_

### Logs

```shell

```

### Notes

_No response_

Botsy:


kKaskak (commented):
## Review  - Opus 4.6

> this review was ai generated

The approach is solid — using a ref to remember user-selected speed and a `skipUpdate` flag to distinguish temporary (hold) vs permanent speed changes is the right pattern. The `forwardRef` for control bar click exclusion is clean.

### Must fix

**`playbackSpeed` ref initializes as `null`** — `video.state.playbackSpeed` is `null` at mount time (see `useVideo.js:23`). If the user holds space before ever manually changing speed, on release it calls `onPlaybackSpeedChanged(null)` instead of restoring `1x`.

```js
const playbackSpeed = React.useRef(video.state.playbackSpeed ?? 1);
```

### Should fix

**`onPlaybackSpeedChanged` deps changed from `[]` to `[video.setPlaybackSpeed]`** — if `video.setPlaybackSpeed` is stable (likely, since it's a `setProp` wrapper), this is a no-op but adds noise. If it's NOT stable, it causes the callback to recreate on every render, breaking memoization downstream since it's passed to `SpeedMenu`. Either:
- Revert to `[]` (it was fine before), or
- Add `onPlaybackSpeedChanged` to the `useLayoutEffect` deps array (`[routeFocused, menusOpen, video.state.volume]`) to avoid the stale closure — right now the effect captures the old callback but `onPlaybackSpeedChanged` can now change identity between renders.

### Looks good

- `SpeedMenu` calls `onPlaybackSpeedChanged(value)` without `skipUpdate` → ref correctly updates when user explicitly picks a speed ✓
- `controlBarRef` guard prevents hold-to-speed on control bar clicks ✓
- `forwardRef` conversion is minimal and clean ✓

kKaskak (approved):

PR #1176, opened 2026-03-26
fix: move streams loading indicator below list to prevent layout shift

### LINKED ISSUE #2187: Polish: When viewing episode video list, when trying to click something the list "jumps"
### Stremio Platform

Windows

### Describe the bug

When you are viewing the list of videos for an episode or film, there is a "Loading" box at the top. When that disappears, which can happen fast or slow, the entire list shifts positions. This has lead to a lot of mis-clicks.

### To Reproduce

View an episode or film video link list

### Expected behavior

If your mouse is in the field, maybe instead of removing the "Loading" box, rename it to "Completed / Finished / Done" so the list doesn't shift. - or display it at the bottom or outside of the list area.

### Screenshots

_No response_

### App Version

5.0.0-beta.30

### Device Info

Windows 10 Pro 9800X3D 3080 TI 64GB DDR5

### OS Version

Windows 10 Pro

### Browser Info

Using the APP

### Logs

```shell

```

### Notes

_No response_

kKaskak:
## Summary
- Moved the "addons are still loading" indicator from above the streams list to below it
- The indicator remains outside the scrollable container so it's always visible
- When it disappears, the stream items above stay in place — no more layout shift causing mis-clicks

Closes Stremio/stremio-bugs#2187

### DIFF
--- a/src/routes/MetaDetails/StreamsList/StreamsList.js
+++ b/src/routes/MetaDetails/StreamsList/StreamsList.js
@@ -170,17 +170,6 @@ const StreamsList = ({ className, video, type, onEpisodeSearch, ...props }) => {
                             </div>
                             :
                             <React.Fragment>
-                                {
-                                    countLoadingAddons > 0 ?
-                                        <div className={styles['addons-loading-container']}>
-                                            <div className={styles['addons-loading']}>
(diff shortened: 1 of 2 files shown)


Botsy (approved):
Changes LGTM, only concern is it may be missed if the addons load too quickly.

PR #1178, opened 2026-03-29
feat(Player): persist subtitle language preference across streams

ignaciojsoler:
When switching to a new episode or title, the selected subtitle language was lost. Now, when the user manually selects a subtitle language in the player, it persists as the global preference and is automatically applied to subsequent titles.


tymmesyde:
Hi, thank you for your contribution
the point of those settings is for them to serve as default value, your implementation breaks this as it modifies the value that the user chose
there is logic already present to save the track id between streams of the same bingeGroup: https://github.com/Stremio/stremio-web/blob/development/src/routes/Player/Player.js#L468
it can also store language, so this could be implemented there

PR #1179, opened 2026-03-29
Player: Remember selected subtitles language 

ignaciojsoler:
Follow-up on #1178  based on feedback from @tymmesyde                                                                                            

Instead of persisting the subtitle language by updating the user's default settings (which would overwrite their configured preference), this implementation follows the suggestion to store the language within the existing streamState mechanism alongside the track ID.

When a subtitle track is selected, its language is now saved in streamState.subtitleTrack.lang. The auto-selection logic then uses this saved language as a fallback when the track ID doesn't match (e.g. next episode in a binge group), before falling back to the user's default subtitlesLanguage setting.

This way the subtitle language is persisted across episodes in the same binge group without modifying the user's configured default.

### DIFF
--- a/src/routes/Player/Player.js
+++ b/src/routes/Player/Player.js
@@ -31,6 +31,7 @@ const { default: Indicator } = require('./Indicator/Indicator');
 
 const findTrackByLang = (tracks, lang) => tracks.find((track) => track.lang === lang || langs.where('1', track.lang)?.[2] === lang);
 const findTrackById = (tracks, id) => tracks.find((track) => track.id === id);
+const getTrackLang = (tracks, id) => id !== null ? findTrackById(tracks, id)?.lang : null;
 
 const Player = ({ urlParams, queryParams }) => {
     const { t } = useTranslation();
@@ -241,9 +242,10 @@ const Player = ({ urlParams, queryParams }) => {
             subtitleTrack: {
                 id,
                 embedded: true,
+                lang: getTrackLang(video.state.subtitlesTracks, id),
             },
         });
-    }, [streamStateChanged]);
+    }, [streamStateChanged, video.state.subtitlesTracks]);
 
     const onExtraSubtitlesTrackSelected = React.useCallback((id) => {
         video.setExtraSubtitlesTrack(id);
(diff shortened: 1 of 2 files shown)


src/routes/Player/Player.js
```diff
@@ -241,19 +242,21 @@ const Player = ({ urlParams, queryParams }) => {
             subtitleTrack: {
                 id,
                 embedded: true,
+                lang: getTrackLang(video.state.subtitlesTracks, id),
```
tymmesyde: here it should only be set with the value of the track that the user selected 
you will need to modify the `onSubtitlesTrackSelected` and `onExtraSubtitlesTrackSelected` functions so that they pass the track as param instead of just the id

src/routes/Player/Player.js
```diff
@@ -466,13 +469,16 @@ const Player = ({ urlParams, queryParams }) => {
             }
 
             const savedTrackId = player.streamState?.subtitleTrack?.id;
+            const savedLang = player.streamState?.subtitleTrack?.lang;
+            const fallbackLang = savedLang ?? settings.subtitlesLanguage;
+
             const subtitlesTrack = savedTrackId ?
                 findTrackById(video.state.subtitlesTracks, savedTrackId) :
-                findTrackByLang(video.state.subtitlesTracks, settings.subtitlesLanguage);
+                findTrackByLang(video.state.subtitlesTracks, fallbackLang);
```
tymmesyde: if  `savedLang` is defined and returned in `fallbackLang` and that the `findTrackByLang` function is not able to find a corresponding track, it will not fallback to `settings.subtitlesLanguage`
so here it should be:

```
                findTrackById(video.state.subtitlesTracks, savedTrackId) :
                findTrackByLang(video.state.subtitlesTracks, savedLang) :
                findTrackByLang(video.state.subtitlesTracks, settings.subtitlesLanguage);
```

src/routes/Player/Player.js
ignaciojsoler: @tymmesyde updated both onSubtitlesTrackSelected and onExtraSubtitlesTrackSelected to receive the full track object, allowing lang to be derived directly from it

src/routes/Player/Player.js
ignaciojsoler: @tymmesyde  simplified the fallback logic by removing fallbackLang and replacing it with a clearer sequence: savedTrackId → savedLang → settings.subtitlesLanguage

Let me know if anything else is needed. Thanks for the feedback!

tymmesyde (approved):

tymmesyde:
Thank you!

PR #1188, opened 2026-04-03
Addons: Add Update Addons functionality

### LINKED ISSUE #196: Addon Upgrade functionality
Add the `Upgrade` button in addon details modal

mrcanelas:
## Summary

This PR updates installed addons **in place** using core `UpgradeAddon`, so addon order is preserved (no uninstall/reinstall flow).

The UX was changed from per-addon update actions to a single **“Update all”** action in the Addons header (installed view), including a mobile-specific icon button layout.

## Related issues
Closes #196 

## What changed

- Added bulk update flow for installed addons:
  - Fetch fresh manifest from `transportUrl`
  - Dispatch `Ctx -> UpgradeAddon` sequentially per addon
  - Correlate core events (`AddonUpgraded` / `Error` with source `AddonUpgraded`) per addon
- Added batch toast muting during "Update all" to avoid per-addon toast spam
- Added protected addon handling:
  - Skip addons flagged as protected in list filtering
  - Treat core `Other / code 5` (`Addon is protected`) as skipped (not failed)
- Improved result reporting:
  - Success: `{count} addons updated`
  - Success: `All addons up to date` when nothing changed and no real failures
  - Partial error: simplified per-addon lines (e.g. `IMDB Catalogs erro: 404`)
- Updated mobile UI:
  - "Update all" appears between search and filter
  - Icon-only button (same shape as filter button), preserving desktop color scheme

## Why

Refreshing addon manifests previously required reinstalling addons, which could change addon order and impact result prioritization.  
Using `UpgradeAddon` in place keeps order stable and makes updates safer and faster for users.

## How to test

1. Open **Addons** (installed tab) with multiple installed addons.
2. Click **Update all** on desktop and mobile layouts.
3. Verify:
   - addon order is unchanged after update;
   - protected addons are not reported as failures;
   - when errors happen, toast shows per-addon simplified lines (e.g. `Name erro: 404`);
   - when all are unchanged/skipped-protected, success message is shown.


tymmesyde:
Hi, can you revert your change where you rewrote the whole component to typescript? 
When opening a PR it's best to stay in scope and modify the least amount of stuff

mrcanelas:
I was told that all the code would be migrated to TS in the future, and I thought I would be helping.

tymmesyde:
The components are rewritten when necessary, if we think they need refactoring or if it's easier for type definitions
But newly added components and other files should be written in typescript

mrcanelas:
> Hi, can you revert your change where you rewrote the whole component to typescript? When opening a PR it's best to stay in scope and modify the least amount of stuff

Done!

kKaskak:
what if we could instead just have one update button on header on the addons page? which would update all addons? wouldnt this be better? I guess some could error out if they are old but we could just show a toast `{number} Addons updated` or `All addons up to date` if none changed

mrcanelas:
> what if we could instead just have one update button on header on the addons page? which would update all addons? wouldnt this be better? I guess some could error out if they are old but we could just show a toast `{number} Addons updated` or `All addons up to date` if none changed

We can do it this way too, just adding one more button instead of one for each add-on. I'll make the changes.

mrcanelas:
> what if we could instead just have one update button on header on the addons page? which would update all addons? wouldnt this be better? I guess some could error out if they are old but we could just show a toast `{number} Addons updated` or `All addons up to date` if none changed

Done!

tymmesyde:
What is the point of this btw? Shouldn't we automatically update the manifests anyway since the server hosting the addon is always running the latest version?

kKaskak:
> What is the point of this btw? Shouldn't we automatically update the manifests anyway since the server hosting the addon is always running the latest version?

auto updating is expensive for our intra, this is a better approach IMO; user takes the action himself if necessary; you can update all or per add-on (during testing, setupping your account etc.)

kKaskak:
@mrcanelas i made the core side implementation and wired the dev package please test it 🙏🏼

kKaskak:
- [ ] update translations 
- [x] remove default value usage

tymmesyde:
>auto updating is expensive for our intra, this is a better approach IMO; user takes the action himself if necessary; you can update all or per add-on (during testing, setupping your account etc.)

is it that different from synchronizing the library? in core we just need to make a request to all the addons and update the old manifests locally, we then synchronize with our API if needed
i don't think the common user know that it could be necessary to update addons and what it does

kKaskak:
> > auto updating is expensive for our intra, this is a better approach IMO; user takes the action himself if necessary; you can update all or per add-on (during testing, setupping your account etc.)
> 
> is it that different from synchronizing the library? in core we just need to make a request to all the addons and update the old manifests locally, we then synchronize with our API if needed i don't think the common user know that it could be necessary to update addons and what it does

ig its a preference to include / exclude manual button

technically it will be only a req to the addons, if all up to date we dont have req to api so its cool

i did the core part so its easy to rewire and do it automatic instead of exposing an action

if @mrcanelas has any reason why he want it to be a Ui button instead of auto update please feel free to share with us

i'm open to discussion on this

only thing im concerned about is that if a dev publishes a new version of the addon which has a bug (removal of a catalog or something similar which does not break the actual manifest; update would fail if it would) and we auto update all users manifests, add-on breaks for everyone. if its manual if a bad release is done users will use the old manifest which will work perfectly fine. and if somebody updates manually only for him the addon breaks and he needs to wait for a fix. 

edit: most users wont reconfigure their addons if everything already works, they arent even visiting the addons page, if something breaks for them without reconfiguring this would lead to online consequences, thats why IMO its better to be manual; they dont even know if a new version is released and what it includes, they are happy with their setup. (excluding users who are active to following addon updates that currently need to uninstall / install addon to update which breaks their addon order)

mrcanelas:
> if @mrcanelas has any reason why he want it to be a Ui button instead of auto update please feel free to share with us

I believe this button is a great addition. Today, with several custom list add-ons thanks to mdblist, every time a new list is added, the manifest changes, requiring reinstallation or the use of tools like [Syncio](https://github.com/iamneur0/syncio), [Stremio Addon Manager](https://github.com/pancake3000/stremio-addon-manager), and others. I think we should have native tools so users don't have to resort to other alternatives, especially simple things like an add-on update button. I don't know how automatic updates work today, if it's a cron job, but it's not instantaneous; when the add-on's manifest changes, it doesn't pass that information to Stremio in real time.

kKaskak:
> > if @mrcanelas has any reason why he want it to be a Ui button instead of auto update please feel free to share with us
> 
> 
> 
> I believe this button is a great addition. Today, with several custom list add-ons thanks to mdblist, every time a new list is added, the manifest changes, requiring reinstallation or the use of tools like [Syncio](https://github.com/iamneur0/syncio), [Stremio Addon Manager](https://github.com/pancake3000/stremio-addon-manager), and others. I think we should have native tools so users don't have to resort to other alternatives, especially simple things like an add-on update button. I don't know how automatic updates work today, if it's a cron job, but it's not instantaneous; when the add-on's manifest changes, it doesn't pass that information to Stremio in real time.
> 
> 
> 
> 

Currently we don't have auto addon update of manifests functionality for users. FYI

mrcanelas:
> Currently we don't have auto addon update of manifests functionality for users. FYI

That's exactly why I believe this manual 'Update All' feature is so relevant right now. Since there is no automated sync for manifests, users are often stuck with outdated catalogs or broken resource links unless they go through the hassle of reinstalling.

This implementation provides a reliable way for users to fetch the latest manifest.json on demand, solving the prioritization issue (maintaining the addon order) without adding the complexity or the overhead of a full background auto-update system at this stage.

src/routes/Addons/Addons.js
```diff
@@ -93,6 +93,23 @@ const Addons = ({ urlParams, queryParams }) => {
...
         platform.openExternal(event.dataset.addon.transportUrl.replace('manifest.json', 'configure'));
     }, []);
+    const hasUpdatableInstalledAddons = React.useMemo(() => {
+        if (installedAddons.selected === null) {
+            return false;
+        }
+        return installedAddons.catalog.some((addon) =>
+            !addon.manifest?.behaviorHints?.configurationRequired && !addon.flags?.protected
+        );
+    }, [installedAddons.selected, installedAddons.catalog]);
```
Botsy, line 103: Is there a way we return `true` only if any of the addons needs to be updated to newer version? Right now we will have this as `true` for any unprotected addon that doesn't need configuration, even if there is no newer version?

Botsy (commented):
@kKaskak @mrcanelas I tested and I honestly think it's a bit confusing why we have the update button if there is no update to an addon. I would suggest we rephrase it a bit into "Check for updates" or something like this. Which will then check if there is an update (and install it ?) and show success message, or show a message that the addon is up-to-date. Is it doable? Otherwise we're showing success messages but doing nothing.

kKaskak:
>  show success message, or show a message that the addon is up-to-date. Is it doable? Otherwise we're showing success messages but doing nothing.

we are already doing this, no? :D

kKaskak:
I agree we could only show the button when we have an update available; we would check against existing manifest vs the one fetched from the addon @Botsy

megaslav:
I think this should be a "Reinstall" button, instead of an "Update" button. 

Sometimes there's no new version of an addon, but it still needs to be reinstalled to reflect changes to catalogs. 

AIOMetadata and AIOStreams even include a message after saving the configuration to tell the user if the changes require the addon to be reinstalled in Stremio in order to be able to see those changes.

It's normal to tinker with the configuration during the setup process, and even every now and then when tastes change, a better idea to organize catalogs pop into your mind, a new addon appears and you want to try it out, etc 

Many users rely on Cinebye, Stremio Addon Manager, AIOManager and similar tools just to reinstall the addon.

AIOStreams:
image

AIOMetadata:
image

mrcanelas:
Updated the wording to better reflect the actual behavior.

- "Update all" → "Check for updates"
- Per-addon action renamed to better represent that it may refresh even when there is no new version
- Adjusted toast messages to distinguish between updated addons and already up-to-date cases

This should avoid the confusion around showing a successful "update" when nothing has actually changed, while keeping the current behavior intact.

PR #1189, opened 2026-04-03
Add touch support for 2x playback

Blade1802:
Hold screen for 2x playback speed like Youtube.

### DIFF
--- a/src/routes/Player/Player.js
+++ b/src/routes/Player/Player.js
@@ -802,12 +802,8 @@ const Player = ({ urlParams, queryParams }) => {
                 }
             }
         };
-
-        const onMouseDownHold = (e) => {
-            if (e.button !== 0) return; // left mouse button only
-            if (menusOpen || nextVideoPopupOpen) return;
-            if (controlBarRef.current && controlBarRef.current.contains(e.target)) return;
-
+    
+        const startHold = () => {
             longPress.current = false;
 
             pressTimer.current = setTimeout(() => {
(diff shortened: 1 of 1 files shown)


kKaskak:
few issues to address:

   ### 1. Touch hold will conflict with native tap-to-play/pause and double-tap-fullscreen

   On touch devices, a `touchstart` fires **before** `click`. After holding for 200 ms the speed changes, but on a *short* tap the browser still synthesizes `click` → `onVideoClick` which
    toggles play/pause. The user taps to pause, but because `longPress.current` is `false` for a quick tap the click handler also runs — that's the same as today's mouse flow and is fine.
    **But** the 200 ms `HOLD_DELAY` is very short on mobile; users commonly hold ≥ 200 ms while scrolling or shifting grip, which will accidentally trigger 2× speed. Consider either:
   - A longer `HOLD_DELAY` for touch (400-500 ms), or
   - Adding a movement threshold: cancel the timer on `touchmove` if the touch moves more than ~10 px (prevents scroll-triggered speed changes).

   ### 2. Missing `touchmove` cancellation

   If the user starts a touch, then scrolls/swipes (e.g., on a seek gesture or volume gesture), the 200 ms timer still fires and activates 2× speed. Add a `touchmove` handler that calls
   `clearTimeout(pressTimer.current)` (similar to how other players handle this).

   ### 3. `{ passive: false }` may be needed

   If you later want to call `e.preventDefault()` in `onTouchStart` (e.g. to prevent the browser's long-press context menu on some Android browsers), you'll need `{ passive: false }`. Not
    blocking for now, but worth noting.

   ### 4. Minor: trailing whitespace

   Several lines have trailing spaces (visible in the diff as whitespace-only changes)

kKaskak (changes requested):
.

PR #1191, opened 2026-04-06
Player: refactor events

kKaskak:
Fixed                                                                                                                                                                                                         
                                         
  - mute (M), subtitlesDelay (G/H), subtitlesSize (-/=), toggleSubtitles (C), speedUp (]), speedDown ([) no longer fire while a menu or popup is open.                                                          
  - StatisticsMenu no longer closes itself when you click inside it.                                                                                                                                            
  - Hold-to-speed-up timer no longer fires after a menu opens mid-press (Space or mouse long-press).                                                                                                            
                                                                                                                                                                                                                
  Refactor                                                                                                                                                                                                      
                                                                                                                                                                                                                
  - onShortcut gained an optional enabled flag so the menu/popup guard lives at the registration site instead of being copy-pasted into each handler.                                                           
  - nextVideoPopupOpen folded into menusOpen — every consumer was already checking both.

### DIFF
--- a/src/common/Shortcuts/onShortcut.ts
+++ b/src/common/Shortcuts/onShortcut.ts
@@ -1,15 +1,16 @@
 import { DependencyList, useCallback, useEffect } from 'react';
 import { ShortcutListener, ShortcutName, useShortcuts } from './Shortcuts';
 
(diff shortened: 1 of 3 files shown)


Botsy (approved):
LGTM

Issue #1144, opened 2026-03-02
[Bug]: Infinite scrolling triggered on few catalog results


SageTendo:
### Stremio-Web Version

/development branch

### Browser

Chrome

### Platform / Device type

Linux

### What Happened?

Infinite scrolling repeatedly requests new pages when returned results are fewer than `PAGE_SIZE`. This can lead to extra requests being made to addons even when no more data exists.

### Logs

```shell

```

### Notes

**Expected Behavior**
- Infinite scroll should stop when the backend has no more/new data.

**Suggestion:**
- Fetch only once on first page load.
- Fetch additional pages only when the user scrolls to the bottom.
- Track the number of items returned in the last fetch and only fetch again if new data was returned:

```javascript
//Instead of triggering scrolling when the content doesn't fill the viewport,
if (containerHeight <= viewportHeight + SCROLL_TO_BOTTOM_THRESHOLD) {
    loadNextPage();
}
```

### Code of Conduct

- [x] I agree

SageTendo:
I can work on this, if that's okay.

sscheetz:
Has anyone repro'd this bug. I would like more clear reproduction steps or maybe a change has already fixed this?

I tried to repro tonight in the following ways

1) Go to discover page and set filters using default cinemeta addon to "Movie", "New", "1920" which only has 11 results. This only started the second row of movies in my viewport.

2) Override the webworker fetch to return a mock response for the 1920 new movie catalog api call with both 0 and 1 potential responses

and was unable to reproduce in either case.

Maybe the specific addon is returning a response that causes this to happen?

kKaskak:
hey created a core pr for this issue 
https://github.com/Stremio/stremio-core/pull/1003

PR #1201, opened 2026-04-19
Discover: Correctly inifinite load results from addons when requested

Sanjays2402:
## Problem

When a catalog addon returns fewer results than `PAGE_SIZE` (100), the auto-fill viewport logic in Discover repeatedly calls `loadNextPage()` because the content never fills the viewport height. This creates an infinite loop of requests to addons even when no more data exists.

Reported in #1144.

## Root Cause

The `useEffect` at lines 39-47 checks if `containerHeight <= viewportHeight + SCROLL_TO_BOTTOM_THRESHOLD` and calls `loadNextPage()` — but never tracks whether the last fetch actually returned new items. When a small catalog can't fill the viewport, it keeps firing.

## Fix

Track the previous item count via a ref (`prevItemsCountRef`). Only trigger auto-fill when new items were actually added since the last check. Reset the counter when the catalog enters `Loading` state (i.e., a new catalog is selected).

This preserves the intended behavior — auto-loading pages until the viewport is filled — while stopping the loop when no new data comes back.

### DIFF
--- a/src/routes/Discover/Discover.js
+++ b/src/routes/Discover/Discover.js
@@ -30,21 +30,30 @@ const Discover = ({ urlParams, queryParams }) => {
 
     const metasContainerRef = React.useRef();
     const metaPreviewRef = React.useRef();
+    const prevItemsCountRef = React.useRef(0);
 
     React.useEffect(() => {
         if (discover.catalog?.content.type === 'Loading') {
             metasContainerRef.current.scrollTop = 0;
+            prevItemsCountRef.current = 0;
         }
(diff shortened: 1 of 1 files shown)


kKaskak:
ok will review soon

kKaskak (approved):

kKaskak:
im not merging this, when testing it appears sometimes it does not load next page even tho there are more results to be loaded needs additional testing

kKaskak:
please fix and i will reopen pr than you ping me

PR #1213, opened 2026-04-20
bump: TypeScript 5 → 6

anikettuli:
## Summary

Part of Phase 1.5 (deferred majors) of the modernization plan.

| Package | Old | New |
|---|---|---|
| typescript | ^5.7.2 | ^6.0.3 |

ts-loader 9.5.x and typescript-eslint 8.x both support TS 6. The repo's
tsconfig uses `strict: true` with `checkJs: false`, so the JS sources
are not type-checked and the blast radius is contained to the ~195
`.ts/.tsx` files.

## Type-check delta

TS 6 introduces deprecation errors for two `tsconfig.json` options that
short-circuit `tsc --noEmit` before it reaches any source-file
type-checking:

```
tsconfig.json(5,9): error TS5101: Option 'baseUrl' is deprecated and will stop functioning in TypeScript 7.0.
tsconfig.json(7,29): error TS5107: Option 'moduleResolution=node10' is deprecated and will stop functioning in TypeScript 7.0.
```

Both are resolvable by adding `"ignoreDeprecations": "6.0"` (or migrating
to `moduleResolution: "bundler"` and removing `baseUrl` in favor of
explicit `paths`). Deferred to Phase 5's TS migration per the
instruction not to modify `tsconfig.json` in this PR.

Once those two config errors are silenced, the underlying source-file
type-check delta can be measured against the documented baseline of 15.
`tsc --noEmit` is not gated in CI (only `pnpm build` + `pnpm test` +
`pnpm lint` run), so this does not block the PR.

## Test plan
- [x] `pnpm install`
- [x] `pnpm lint`
- [x] `pnpm test` (70 passed)
- [x] `pnpm build` (compiled with 3 unrelated size-warning messages; `build/*/scripts/main.js` and `build/*/scripts/worker.js` emitted)

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
         "terser-webpack-plugin": "5.3.10",
         "thread-loader": "^4.0.4",
         "ts-loader": "^9.5.1",
-        "typescript": "^5.7.2",
+        "typescript": "^6.0.3",
         "typescript-eslint": "^8.17.0",
         "webpack": "5.97.0",
         "webpack-cli": "5.1.4",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -52,7 +52,7 @@ importers:
         version: 0.0.3
       i18next:
         specifier: ^24.0.5
-        version: 24.2.3(typescript@5.9.2)
+        version: 24.2.3(typescript@6.0.3)
       langs:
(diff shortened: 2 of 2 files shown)


anikettuli:
Superseded by #1214 — rolled up into a single consolidated dependency-refresh PR.

PR #1216, opened 2026-04-20
chore: add Stylelint for .less files, wire into CI

anikettuli:
Adds a Stylelint config + npm script + CI step for the 117 Less files in `src/`. Current state: the existing Less is clean under the minimal config introduced here, minus four small real issues that are fixed in the same commit.

## Config (`.stylelintrc.cjs`)

- Extends **`stylelint-config-recommended`** (catches syntax errors, unknown standard CSS properties, duplicate keyframe names, etc.).
- Uses **`postcss-less`** as the custom parser.
- Relaxes rules that are intentionally violated by the repo's existing style:

| Relaxation | Why |
|---|---|
| `:global`, `:local`, `:import` allowed in pseudo-class list | CSS Modules syntax |
| `at-rule-no-unknown` / `at-rule-prelude-no-invalid` off | `@import (reference)`, Less at-rules, CSS Modules `composes:` |
| `property-no-unknown` off | CSS Modules `composes` declarations read as unknown properties |
| `media-query-no-invalid` off | Less variables inside media queries (`(max-width: @minimum)`) |
| `no-descending-specificity` / `no-duplicate-selectors` off | `:global(.selected)` prefixed selectors naturally create both |
| `color-hex-length` / `color-function-notation` / `alpha-value-notation` off | Repo uses short + long hex and mixed color notations throughout |
| `function-no-unknown` off | Less `@{var}` interpolation in functions |
| `declaration-block-no-duplicate-properties` off | Intentional fallbacks (`display: -webkit-box` then `display: flex`) |
| `block-no-empty` off | Empty rule blocks used as composition placeholders |
| `declaration-property-value-no-unknown` off | Less variables in `calc()` and property values (and it hides a few legit issues, but those aren't blocking adoption) |

## Real issues fixed while tuning the config

The first pass surfaced **4 legitimate issues** that got cleaned up instead of silenced:

| File | Change |
|---|---|
| `src/App/styles.less:130` | `word-break: break-word` (deprecated) → `overflow-wrap: break-word` (modern equivalent, same visual behaviour). |
| `src/App/styles.less:160` | `'TwemojiFlags', 'sans-serif'` → `'TwemojiFlags', sans-serif` (the quoted `'sans-serif'` treats the generic as a literal family name, which is a bug — browsers then fall back to the default). |
| `src/routes/Discover/styles.less:180` | `grid-gap: 0.5rem` → `gap: 0.5rem` (modern standard property name). |
| `src/routes/Library/styles.less:96` | Same `grid-gap` → `gap`. |

## Dev deps

| Package | Version |
|---|---|
| `stylelint` | `^17.8.0` |
| `stylelint-config-recommended` | `^18.0.0` |
| `postcss-less` | `^6.0.0` |

## New script

```
pnpm lint:styles
```

## CI

`.github/workflows/build.yml` gains a `Lint styles` step immediately after `Lint`, so style regressions block PR merges from now on.

## Known follow-ups (not fixed here)

While relaxing `declaration-property-value-no-unknown`, a handful of suspicious values were also ignored:
- `font-size: 500` in `src/routes/Calendar/Placeholder/Placeholder.less:59` and `src/routes/Library/Placeholder/Placeholder.less:59` — `500` looks like a misplaced `font-weight`.
- `background-color: none` in `src/routes/MetaDetails/StreamsList/styles.less:118` — `none` isn't valid for `background-color`; probably should be `transparent`.

These are left alone to keep the scope of this PR to "introduce the lint gate without changing behaviour"; a follow-up can tighten the rule and fix them.

## Test plan

- [x] `pnpm install` — clean
- [x] `pnpm lint:styles` — 0 errors
- [x] `pnpm lint` — clean
- [x] `pnpm test` — 70/70 across 3 suites
- [x] `pnpm build` — webpack compiles, only pre-existing asset-size warnings

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,6 +39,8 @@ jobs:
         run: pnpm test
       - name: Lint
         run: pnpm lint
+      - name: Lint styles
+        run: pnpm lint:styles
       # Create recursively the destination dir with
       # "--parrents where no error if existing, make parent directories as needed."
       - run: mkdir -p ./build/${{ github.head_ref || github.ref_name }}
--- a/.stylelintrc.cjs
+++ b/.stylelintrc.cjs
@@ -1,0 +1,51 @@
+// Copyright (C) 2017-2026 Smart code 203358507
+
+/** @type {import('stylelint').Config} */
+module.exports = {
+    extends: ['stylelint-config-recommended'],
+    customSyntax: 'postcss-less',
+    rules: {
+        // CSS Modules :global, :local, :import pseudo-classes.
+        'selector-pseudo-class-no-unknown': [true, {
+            ignorePseudoClasses: ['global', 'local', 'import'],
+        }],
+        // CSS Modules `composes:` and `@value`, Less syntax.
+        'at-rule-no-unknown': null,
+        'at-rule-prelude-no-invalid': null,
+        'property-no-unknown': null,
+        // Less variables appear inside media queries (@{minimum} etc.).
+        'media-query-no-invalid': null,
+        // Less tilde imports (~pkg/file.less) and `@import (reference)`
+        // are valid in the webpack + less-loader setup.
+        'no-invalid-position-at-import-rule': null,
+        // :global(...) wrappers naturally create specificity inversions
+        // and duplicate selectors compared with the scoped local class.
+        'no-descending-specificity': null,
+        'no-duplicate-selectors': null,
+        // The repo uses short-hex (#fff) and long-hex (#ffffff)
+        // interchangeably; mixed notation is intentional.
+        'color-hex-length': null,
+        'color-function-notation': null,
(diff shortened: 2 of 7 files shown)


kKaskak:
what does this solve

anikettuli:
117 `.less` files with no linter gating them. While tuning the config, it caught 4 real bugs (all fixed in the PR): deprecated `word-break: break-word` in `App/styles.less`, a quoted `'sans-serif'` generic that silently broke font fallback, and two `grid-gap` → `gap` updates. Config is minimal (`stylelint-config-recommended` only, not `-standard`) and tuned for the CSS Modules + Less patterns the repo already uses. No formatting rules, no forced auto-fix, just catches the bug classes ESLint can't see.

PR #1218, opened 2026-04-21
Discover: Add native EPG support to Live TV

mrcanelas:


src/routes/Discover/EpgGuide/EpgGuideRow/EpgGuideRow.tsx
Botsy: It would be better to make the current program highlight and "now" line to update regularly. Right now they update only on refresh or on some user interaction. I would suggest setting an interval like this:
```
const [now, setNow] = useState(Date.now);
    useEffect(() => {
        const id = setInterval(() => setNow(Date.now()), 60000);
        return () => clearInterval(id);
    }, []);
```

src/routes/Discover/EpgGuide/EpgGuide.tsx
```diff
@@ -156,24 +178,55 @@ const EpgGuide = ({ addon, onProgramSelect }: Props) => {
         }
     }, [effectiveDay, pixelsPerHour]);
 
-    const handleSlotSelect = useCallback((index: number) => {
+    const handleSlotSelect = useCallback((value: string | number | null) => {
+        const index = Number(value);
+        if (!Number.isInteger(index)) return;
```
Botsy, line 181: We need to guard against `null` value too. Otherwise, if value is `null`, 00:00 will be selected.

Botsy (changes requested):
I tested and it works but there are many things for improvement on the UX side, major ones are:

- When the user clicks on a show they see the info for a particular show but then if they open the stream, it always loads the currently running show. It might be confusing for the user, especially for shows that already have passed. We need to come up with a solution how to notify/inform the user that streams for passed shows (for now?) and shows in the future are not available and they could only watch in real time.
- When the user opens a channel stream the seekbar progress looks weird and we probably should hide it for live TV and show "LIVE" label or something that indicates the user cannot seek back/forward
- if the user starts watching a channel and goes back to the show meta details, there is a progress bar showing which is misleading in this case

Botsy:
@mrcanelas can you please resolve conflicts as well?

src/routes/Discover/EpgGuide/EpgGuideRow/EpgGuideRow.tsx
mrcanelas: Done. I moved the current time tracking to the Discover EPG flow and refresh it every 60 seconds while the EPG layout is active. The `now` value is now passed down to the guide and also reused by the EPG preview modal, so the current program highlight, the "now" line, and the stream availability check stay in sync without requiring a refresh or user interaction.

src/routes/Discover/EpgGuide/EpgGuide.tsx
mrcanelas: Done. Added an explicit `null` guard before converting the selected slot value to a number, so clearing/empty selection no longer falls back to `Number(null) === 0` and accidentally selects `00:00`.

mrcanelas:
> I tested and it works but there are many things for improvement on the UX side, major ones are:
> 
> * When the user clicks on a show they see the info for a particular show but then if they open the stream, it always loads the currently running show. It might be confusing for the user, especially for shows that already have passed. We need to come up with a solution how to notify/inform the user that streams for passed shows (for now?) and shows in the future are not available and they could only watch in real time.
> * When the user opens a channel stream the seekbar progress looks weird and we probably should hide it for live TV and show "LIVE" label or something that indicates the user cannot seek back/forward
> * if the user starts watching a channel and goes back to the show meta details, there is a progress bar showing which is misleading in this case

Thanks, agreed.

The first point is addressed in this PR now: the EPG preview still opens for past/future programs so users can read the program info, but the stream/show action is only available for the currently airing program. This avoids sending users from a past/future EPG item to the live channel stream for a different program.

For the seekbar/progress behavior in the player and meta details, I agree those are valid UX issues, but they are broader live TV/player behavior issues that existed before this EPG implementation. I think it would be better to handle them in separate PRs so this one stays focused on adding the EPG guide and its direct interactions.

Botsy:
@mrcanelas I tested your changes locally and all you did is awesome and in the right direction. 🙌 A few things I noticed that could be further improved:
- Can we make the "next video popup" appear shortly before the start of the next show in the program? I believe you already mentioned this as possible improvement.
- I would still suggest hiding the progress bar or tweaking it to look in a way that makes it clear it's live show.
- Right now the watched TV channel is being added with progress to the "Continue watching" section on the Board which looks weird. Probably we could detect if it's from EPG addon to not show the progress at all?
- Same issue with progress bar you can see on Meta details page for a channel you already watched and it looks misleading. Maybe we should also hide the progress bar there, too.

Botsy (approved):
@mrcanelas I am merging so that I can apply needed core contract changes on top of yours here

PR #1237, opened 2026-04-28
App: Correctly display fullscreen status

AKnassa:
## TL;DR

`useFullscreen` previously held its `fullscreen` boolean in component-local `useState(false)`. Each consumer (`HorizontalNavBar`, `NavMenuContent`, `Player`) owned an independent copy, so route changes that remounted a consumer caused its local state to default back to `false` even when the document was *actually* fullscreen — because `fullscreenchange` only fires on real transitions, not on remounts.

This PR lifts fullscreen state into a single `FullscreenProvider` mounted at the app root. All consumers read from one context value, the source of truth is initialized from `document.fullscreenElement`, and settings are kept fresh via the `SettingsUpdated` core event.

## Architecture: before vs after

ig_0e2d044caae910930169f3ac5b87fc819b98d115a044fb62b9

## The bug
If you went fullscreen and then switched to another tab in the app, the fullscreen icon button in the nav bar silently stopped working. Clicking it did nothing — there was no way to exit fullscreen from the UI without reloading the page.

### Reproduction (pre-fix)
1. Open the app on `/board`.
2. Press `F` to enter fullscreen — the nav-bar fullscreen icon flips to "exit fullscreen". ✅
3. Click into `/discover` (or any other tab). The route swaps; `HorizontalNavBar` remounts.
4. The icon now shows "enter fullscreen" even though the browser is fullscreen.
5. Pressing the icon (or `F`) calls `requestFullscreen` again — a no-op in some browsers, a `TypeError` in others; UI and reality stay out of sync until the user manually exits and re-enters.

bug

### Why it happened

```js
// old useFullscreen
const [fullscreen, setFullscreen] = useState(false);
useEffect(() => {
  document.addEventListener('fullscreenchange', onChange);
  return () => document.removeEventListener('fullscreenchange', onChange);
}, []);
```

Three problems compound:
1. **Local state, multiple consumers.** Three components each call `useFullscreen()`, each gets its own `[fullscreen, setFullscreen]`. They never share.
2. **`useState(false)` ignores reality.** On mount, the hook asserts "we are not fullscreen" regardless of `document.fullscreenElement`.
3. **`fullscreenchange` only fires on transitions.** A freshly mounted listener will *not* receive a synthetic event for the current state — it just sits at `false` until the next real toggle.

A route change that unmounts a consumer and mounts a sibling triggers all three. Result: state desync.

### 1. Single provider, single state

A new `FullscreenProvider` is mounted once at the app root, above the router. State lives there and is exposed via `FullscreenContext`:

```tsx
const [fullscreen, setFullscreen] = useState<boolean>(() => {
  if (typeof document === 'undefined') return false;
  return document.fullscreenElement === document.documentElement;
});
```

Two things matter here:
- The provider is **above** the router, so it doesn't unmount on navigation. Its state survives every route change.
- The `useState` lazy initializer reads the *actual* fullscreen state on first render. This handles the case where the provider is created while the document is already fullscreen (e.g. hard refresh, deep-link into a fullscreen-restored session).

### 2. Three consumers now read context, not local state

`HorizontalNavBar`, `NavMenuContent`, and `Player` were updated to consume `FullscreenContext` via `useFullscreen()`. Their per-instance `useState` and event listeners are gone — there's exactly one source of truth in the app.

### 3. Settings access without `useModelState`

The provider needs `escExitFullscreen` (a profile setting) so `Esc` can exit fullscreen when the user has opted in. Previously, hooks like `useSettings()` were used — but those rely on `useModelState` → `useCoreSuspender`, which only works **beneath** the router's `withCoreSuspender` boundary. Since `FullscreenProvider` sits above the router, those hooks aren't available.

So the provider talks to `core.transport` directly:
- **Initial seed** — `getState('ctx')` once on mount, picks `profile.settings.escExitFullscreen`.
- **Live updates** — listens on `CoreEvent` and filters for `SettingsUpdated`, reading `escExitFullscreen` straight from the event payload.

This is the same pattern `App.js` already uses for `interfaceLanguage` and `quitOnClose`. No new trust boundary, no new transport plumbing.

### 4. Why `CoreEvent` / `SettingsUpdated` and not `NewState`

An earlier draft used `NewState` — it fires when any model changes. Problem: the `ctx` model also covers notifications, search history, library sync, and streaming-server URL. Subscribing to `NewState` and filtering for `'ctx'` meant a `getState('ctx')` postMessage round-trip on every notification poll, every search, every library sync. The provider was doing real work to re-read a setting that almost never changes.

The final commit (`perf(fullscreen): subscribe to SettingsUpdated, not every ctx NewState`) switches to `CoreEvent` filtered to `SettingsUpdated`, so we only react when settings *actually* change, and we read the new value straight from the event args — no extra round-trip.

## What changed

| Area | Change |
|---|---|
| `src/common/Fullscreen/FullscreenContext.ts` | New context module. Tuple-shaped value `[fullscreen, requestFullscreen, exitFullscreen, toggleFullscreen]` to keep the call-site identical to the old hook. |
| `src/common/Fullscreen/FullscreenProvider.tsx` | New provider. Owns state, owns DOM listeners (`fullscreenchange`, `keydown`), owns shell `win-visibility-changed` subscription, owns settings read. |
| `src/common/Fullscreen/useFullscreen.ts` | Now a thin `useContext` wrapper — same return shape as before, drop-in replacement. |
| `src/App/App.js` | `<FullscreenProvider>` mounted inside `<ShortcutsProvider>` (so shortcut keymap is available) and above the router. |
| `src/components/NavBar/HorizontalNavBar/*` | Removed local fullscreen state; reads from context. |
| `src/components/NavBar/HorizontalNavBar/NavMenu/NavMenuContent.js` | Same. |
| `src/routes/Player/Player.js` | Same. |

Each piece landed in a separate commit so the history tells a clean story:

```
60df6860d  feat(common): add FullscreenProvider + context module
c97a9f156  feat(app): mount FullscreenProvider at app root
b7f7a3d2e  fix(fullscreen): consume FullscreenProvider, remove per-instance state
2e13a6000  fix(fullscreen): read settings via core.transport, not useSettings
90e2cbff1  chore(fullscreen): drop ts-expect-error, type ctx + NewState properly
d3d35bcb4  perf(fullscreen): subscribe to SettingsUpdated, not every ctx NewState
```

## Behavior change worth flagging

**The `F` shortcut now toggles fullscreen on every route.** Previously, the keydown listener was installed inside `useFullscreen()`, which meant it only existed where a consumer was mounted — `HorizontalNavBar`, `NavMenuContent`, or `Player`. Now it's installed once in the provider at the app root, so `F` works on routes that previously ignored it (e.g. `Intro`, modal-only flows, `Settings`).

This is almost certainly the desired behavior — having the shortcut work inconsistently across routes was an accident, not a feature — but it's a user-visible change. QA should spot-check that `F` doesn't conflict with anything on those routes (notably `Intro`, `Settings`, and modal-heavy flows).

`F11` in the desktop shell is unchanged. `Esc` still requires `escExitFullscreen=true` to bail out of fullscreen.

## Trade-offs and what was deliberately not done

- **Default context value uses noops, not "throw if outside provider".** The provider wraps the entire router, so any `useFullscreen()` call has a real value. A throwing default would catch a misuse that can't currently happen; today it'd just produce friendly-fire crashes.
- **`escExitFullscreen` lives in a `ref`, not state.** The keydown handler reads it through `.current`; a re-render isn't needed on settings change. Storing it in state would cause every consumer to re-render whenever settings update, for no observable benefit.
- **`Esc` is not gated by `inputFocused`.** This matches the pre-fix behavior. With `escExitFullscreen=true`, pressing `Esc` to clear a search input also exits fullscreen. It's mildly annoying, but it's not a regression introduced by this PR. Out of scope; tracking separately.
- **No retries/back-off on `getState('ctx')` failures.** A failed initial read leaves `escExitFullscreenRef = false`, which means `Esc` won't exit fullscreen on first load if the worker is wedged. Acceptable: the worker being wedged is a much bigger problem than this single feature, and `SettingsUpdated` will refresh the value on the next user-driven settings change.

## Test plan

Manual QA — please cover at least the following:

- [ ] **Regression check.** Toggle fullscreen with \`F\` on Board / Discover / Library, then navigate to another route. Fullscreen state stays correct (icon, behavior).
- [ ] **Three-way consistency.** Toggle from Player double-click + \`F\`-key + nav-bar fullscreen button. All three reflect a single state.
- [ ] **Esc with \`escExitFullscreen=on\`.** Enter fullscreen, press Esc → exits.
- [ ] **Esc with \`escExitFullscreen=off\`.** Enter fullscreen, press Esc in Player → \`onShortcut('exit')\` (history.back) still fires; fullscreen stays.
- [ ] **Live settings update.** While fullscreen, toggle \`escExitFullscreen\` in Settings. Next \`Esc\` honors the new value (validates the \`SettingsUpdated\` subscription).
- [ ] **F-key behavior change.** Confirm \`F\` now toggles fullscreen on Intro, Settings, and modal-only routes — and that it doesn't conflict with existing shortcuts there.
- [ ] **PWA path.** Browser, no shell — \`requestFullscreen\` / \`exitFullscreen\` drives \`document.documentElement\`.
- [ ] **Stremio shell path.** Native shell active — \`F11\` and the toggle send \`win-set-visibility\` to the shell.
- [ ] **Strict-Mode dev build.** No duplicate \`keydown\` handlers (one toggle per \`F\` press, not two).

## Security

No findings.

- No new network calls, no new persistence, no DOM injection / `innerHTML`, no `eval`.
- Settings (`escExitFullscreen`) read via the existing in-process `core.transport`; no new trust boundary crossed.
- `console.error` only logs failure strings, never settings or user data.
- `KeyboardEvent.code` is read but only compared to literals (`'Escape'`, `'KeyF'`, `'F11'`); never used as a selector or dispatched.
- The `Ctx` cast on `getState('ctx')` is type-only; the value is read through optional chaining and coerced with `!!`, so a schema drift produces `false`, not a crash.


kKaskak (changes requested):
F key shortcut exists in two independent systems

shortcuts.json declares a "fullscreen" shortcut mapped to F, which ShortcutsProvider matches and dispatches. But no component calls onShortcut('fullscreen', handler) — the actual toggle
lives in FullscreenProvider's own keydown listener. Both fire on every F press; the shortcuts one is dead dispatch.

AKnassa:
Good catch, thanks. Pushed 35b100767 — F now subscribes via onShortcut('fullscreen', toggleFullscreen, [toggleFullscreen]) and the duplicate KeyF branch is gone, so the canonical shortcut dispatch is no longer dead.

I left Escape and F11 in the local handler on purpose: Escape is gated on the escExitFullscreen profile setting (which the shortcuts system doesn't express), and F11 is shell-only and not declared in shortcuts.json. Happy to fold either of them in if you'd prefer, let me know.

kKaskak:
@AKnassa please resolve conflicts

AKnassa:
@kKaskak - thanks for your patience on this! I've resolved all the merge conflicts. Could you please review the updates when you get a chance and merge once it looks good to you? Really appreciate the review!

kKaskak:
@AKnassa please remove the comments in the code they are are unnecessary, please update copyright in new files to 2026 correctly

AKnassa:
@kKaskak - Thanks for the review! Addressed both in dfe0d08a7; removed the rationale comments and bumped the copyright to 2026 on the four new files. Let me know if you'd like anything else tweaked.

src/common/Fullscreen/FullscreenProvider.tsx
```diff
@@ -0,0 +1,127 @@
...
+                document.exitFullscreen();
+            }
+        }
+    }, [shell]);
+
+    const toggleFullscreen = useCallback(() => {
+        fullscreen ? exitFullscreen() : requestFullscreen();
+    }, [fullscreen, exitFullscreen, requestFullscreen]);
+
+    onShortcut('fullscreen', toggleFullscreen, [toggleFullscreen]);
```
kKaskak: This moves F from the old hook keydown handler, which skipped INPUT, TEXTAREA, SELECT, and contentEditable targets, into ShortcutsProvider. ShortcutsProvider dispatches while text fields are focused, so typing f in Search, Intro credentials, or settings inputs now toggles fullscreen. Please keep the old activeElement guard around this listener or add an equivalent shortcut-level guard.

kKaskak (changes requested):

src/common/Fullscreen/FullscreenProvider.tsx
AKnassa: @kKaskak - Thanks for catching this

I restored the editable-target guard around the fullscreen shortcut, so `F` is ignored while an `INPUT`, `TEXTAREA`, `SELECT`, or `contentEditable` element has focus.

Tested:
• `F` toggles fullscreen when normal page content is focused.
• Search input accepts `f` / `F` without toggling fullscreen.
• Intro email/password fields accept `f` without toggling fullscreen.
• Settings Streaming “Add URL” input accepts `f` without toggling fullscreen once focused.
• Fullscreen button, Escape exit, and the existing non-input fullscreen flow still behave correctly.

kKaskak (dismissed):

kKaskak:
@AKnassa you have the core loaded in fullscreen provider cant you use it as well with the correct hooks? instead of using the raw approach? check other contexts; except im missing something

AKnassa:
@kKaskak  - Thank you for the approval earlier. But yes having hook is cleaner here, so I updated this to use the existing hook path instead of manually reading/parsing core payloads.

`FullscreenProvider` is now exported through `withCoreSuspender`, so it can safely use `useSettings()` for `escExitFullscreen`. I kept the provider as the single shared fullscreen state/listener owner, and kept the focused input/contentEditable guard for the fullscreen shortcut.

I also updated the architecture image in the PR description; please see that for context.

kKaskak (dismissed):

kKaskak (approved):

kKaskak:
thank you for your contribution

PR #1244, opened 2026-04-30
Dev: Offload all subtitles logic from Player to dedicated hook

kKaskak:


Botsy (approved):
Tested, all looks good 👍

PR #1245, opened 2026-04-30
feat: add playback segment support (intro & credits)

chaffy999:
### 🎯 Vision / Objective

This PR introduces a **generic playback segments system** in Stremio Web, enabling time-based UI interactions during video playback.

The initial implementation supports:

* **Intro segments → Skip Intro overlay**
* **Credits segments → Early Next Episode trigger window**

The goal is not to implement a feature-specific behavior, but to establish a **clean separation of responsibilities between metadata providers and the UI layer**.

---

### 🧠 Architectural Vision: Clear Separation of Concerns

This change reinforces a strict separation between:

#### 🔌 Addons (Data Providers)

* Responsible for providing **metadata per stream**
* May include optional `behaviorHints.segments`
* Can define:

  * intro ranges
  * credits ranges
  * or any future segment types
* Fully flexible: users can implement or choose their own addons / APIs

👉 Stremio does **not enforce how this data is generated**

---

#### 🖥️ Stremio Web (UI / Decision Layer)

* Responsible for:

  * interpreting segment metadata
  * deciding when to show overlays
  * handling user interactions (Skip Intro, Next Episode triggers)
* Implements UI logic only
* Does NOT assume correctness of metadata

👉 Stremio acts purely as a **rendering and orchestration layer**

---

### 📦 Data Model Introduced

```json
{
  "behaviorHints": {
    "segments": [
      { "type": "intro", "start": 52, "end": 110 },
      { "type": "credits", "start": 2400, "end": 2600 }
    ]
  }
}
```

This structure is:

* optional
* extensible
* fully addon-driven

---

#### 1. Skip Intro (UI-driven)

* Detects active `intro` segment during playback
* Displays “Skip Intro” button
* Seeks to segment end on interaction

---

#### 2. Credits-aware Next Episode triggering

Credits segments are used as a **temporal extension of the Next Episode system**:

* When playback enters a `credits` segment:

  * Stremio Web may trigger **Next Episode overlay earlier than end-of-video**
* If no next episode exists:

  * no additional UI is forced (keeps existing behavior)

👉 This enhances continuity without changing core playback logic.

---

### 🔄 Key Principle: Stremio is not responsible for metadata correctness

This design ensures:

* Stremio does NOT validate or enforce segment correctness
* Any discrepancies (wrong timestamps, missing data, incorrect detection) are **fully delegated to the addon or external provider**
* UI behaves strictly based on provided data

👉 This avoids coupling Stremio to:

* external APIs (e.g. IntroDB-like services)
* community-generated metadata quality
* provider-specific inconsistencies

---

### 🌐 Extensibility Benefit

This architecture enables users and developers to:

* plug in **any metadata provider they want**
* experiment with different segment detection systems
* build custom APIs without modifying Stremio Web
* iterate independently on:

  * intro detection
  * credit detection
  * chapter systems

👉 Stremio becomes a **neutral execution layer**, not a data authority.

---

### ⚙️ Behavior Rules

* No segment data → default behavior unchanged
* Invalid segments → safely ignored
* Overlapping segments → credits takes priority over intro
* Seeking behavior → recalculates active segment state
* Next Episode logic remains unchanged except for credits early trigger window

---

### 🚀 Impact

This PR establishes a foundation for:

* chapter-like playback experiences
* richer binge-watching flows
* future segment types (recap, chapters, etc.)
* full decoupling between metadata sources and UI behavior

---

### 💡 Summary

Stremio Web now supports a **provider-agnostic playback segment system**, where:

* Addons define *what exists*
* Stremio Web defines *how it is presented*

This ensures maximum flexibility, extensibility, and independence from any single metadata source or detection strategy.

---

The translations PR is available here:
[https://github.com/Stremio/stremio-translations/pull/1041](https://github.com/Stremio/stremio-translations/pull/1041)

### DIFF
--- a/src/routes/Player/Player.js
+++ b/src/routes/Player/Player.js
@@ -16,6 +16,11 @@ const VolumeChangeIndicator = require('./VolumeChangeIndicator');
 const Error = require('./Error');
 const ControlBar = require('./ControlBar');
 const NextVideoPopup = require('./NextVideoPopup');
+const SegmentActionOverlay = require('./SegmentActionOverlay');
+const {
+    normalizePlaybackSegments,
+    findActivePlaybackSegment,
+} = require('./playbackSegments');
 const StatisticsMenu = require('./StatisticsMenu');
 const OptionsMenu = require('./OptionsMenu');
 const SubtitlesMenu = require('./SubtitlesMenu');
@@ -77,9 +82,28 @@ const Player = ({ urlParams, queryParams }) => {
     const [nextVideoPopupOpen, openNextVideoPopup, closeNextVideoPopup] = useBinaryState(false);
     const [sideDrawerOpen, , closeSideDrawer, toggleSideDrawer] = useBinaryState(false);
 
+    const normalizedPlaybackSegments = React.useMemo(() => {
+        return normalizePlaybackSegments(
+            player.selected !== null ? player.selected.stream : null,
+        );
+    }, [player.selected]);
+
+    const activePlaybackSegment = React.useMemo(() => {
+        if (video.state.time === null) {
+            return null;
+        }
+        return findActivePlaybackSegment(
+            video.state.time,
+            normalizedPlaybackSegments,
+        );
+    }, [video.state.time, normalizedPlaybackSegments]);
+
+    const skipIntroOverlayOpen =
+        activePlaybackSegment !== null && activePlaybackSegment.type === 'intro';
+
     const menusOpen = React.useMemo(() => {
-        return optionsMenuOpen || subtitlesMenuOpen || audioMenuOpen || speedMenuOpen || statisticsMenuOpen || sideDrawerOpen || nextVideoPopupOpen;
-    }, [optionsMenuOpen, subtitlesMenuOpen, audioMenuOpen, speedMenuOpen, statisticsMenuOpen, sideDrawerOpen, nextVideoPopupOpen]);
+        return optionsMenuOpen || subtitlesMenuOpen || audioMenuOpen || speedMenuOpen || statisticsMenuOpen || sideDrawerOpen || nextVideoPopupOpen || skipIntroOverlayOpen;
+    }, [optionsMenuOpen, subtitlesMenuOpen, audioMenuOpen, speedMenuOpen, statisticsMenuOpen, sideDrawerOpen, nextVideoPopupOpen, skipIntroOverlayOpen]);
 
     const closeMenus = React.useCallback(() => {
         closeOptionsMenu();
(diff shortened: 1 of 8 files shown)


kKaskak:
hi i am thankful for your contribution, however i will need to reject this PR. This is because the Stremio vision of this feature differs from what you did in your PR. the biggest reason being that users will be very confused about why some streams have this why other dont. a normal user does not know what info is coming from the addon. he will just ask why it is not working why dont i get skip intro outro. and for every addon to support this means each developer would need to add this field to his addon and then get the data from somewhere? scrape or api based for each stream? this is a lot of work and it would be repetetive across all addons

Regardless, if you would like to contribute in the future with diff prs please read the contribution guide and whats the structure of Stremio Web/Desktop in general

chaffy999:
Hi, thanks a lot for the feedback. I understand the concern about UX consistency and addon fragmentation.

I would be really interested to better understand what the envisioned approach is on your side for implementing a "Skip Intro / Credits" feature in a way that aligns with Stremio’s architecture.

From my perspective, there are a couple of constraints that make this problem tricky:

* In Stremio’s streaming model with multiple addons and heterogeneous HTTP or torrent streams, doing reliable client side detection seems quite difficult without heavy processing or inconsistent results depending on the device.
* For client side detection to be reasonably accurate, you often need some form of cross episode comparison such as visual similarity, audio fingerprinting, or recurring intro patterns. This would require access to multiple episodes ahead of time, which does not naturally fit with how streams are currently resolved and played.
* Users are already exposed to variability today such as stream availability or quality differences, so some level of feature availability depending on the stream already exists in practice.

The approach I proposed was mainly trying to:

* keep Stremio clients focused on UI and UX only
* allow external providers such as addons or APIs to supply segment data
* make the system extensible, whether through ML based APIs or community driven datasets, without hardcoding a single source into Stremio

Regarding the concern that every addon developer would need to implement this, one possible alternative could be to move this kind of metadata to a higher level, for example at the movie or episode metadata layer instead of the stream level. This could allow a single dedicated addon to provide segment data for all content, rather than requiring each stream addon to implement it independently.

That said, this approach would also mean that there is no reliable way to perform precise detection based on the actual file or stream being played, and the feature would therefore be inherently less accurate by design. It would be a tradeoff between ease of integration and precision.

So I am curious:
Is the plan to have a centralized Stremio managed solution such as an internal API, dataset, or detection pipeline?
Or do you envision a client side detection approach?
Or something else entirely?

I would like to align with the long term vision before iterating further, especially since this feature touches both UX expectations and architecture quite deeply.

Also, if there are plans to move forward with this feature, I would be very interested in helping with the implementation or contributing where it makes sense.

Thanks again for the review

PR #1252, opened 2026-05-02
Player: Improve handling of macOS media controls

erik-kroon:
## Summary

- handle explicit shell `media-key` `play` and `pause` actions in the player
- route Media Session `previoustrack` through the same previous-track behavior as shell media keys
- keep Media Session playback state and action handlers cleaned up when the player unmounts

## Why

macOS/AirPods controls can arrive as explicit `play` or `pause` events instead of only `play-pause`. The shell handler previously ignored those explicit actions, so hardware or system media controls could fire without changing playback.

## Behavior

- `play` calls the existing player play callback
- `pause` calls the existing player pause callback
- `play-pause` toggles only when playback state is known
- `next-track` keeps advancing to the next video when available
- `previous-track` keeps seeking to the beginning after 5 seconds
- Media Session handlers are reset on cleanup

## Validation

- `pnpm lint`
- `pnpm build`

Build completes with the existing webpack asset-size/performance warnings.

## Manual macOS checks

- AirPods remove-to-pause
- AirPods stem/button play/pause
- keyboard media play/pause
- macOS Control Center play/pause
- focused and unfocused app states

### DIFF
--- a/src/routes/Player/Player.js
+++ b/src/routes/Player/Player.js
@@ -255,6 +255,12 @@ const Player = ({ urlParams, queryParams }) => {
         }
     }, [player.nextVideo, handleNextVideoNavigation, profile.settings]);
 
+    const onPreviousTrackRequested = React.useCallback(() => {
+        if (video.state.time !== null && video.state.time > 5000) {
+            onSeekRequested(0);
+        }
+    }, [video.state.time, onSeekRequested]);
+
     const onVideoClick = React.useCallback(() => {
         if (video.state.paused !== null && !longPress.current) {
             if (video.state.paused) {
@@ -534,13 +540,21 @@ const Player = ({ urlParams, queryParams }) => {
         }
     }, [settings.pauseOnMinimize, shell.windowClosed, shell.windowHidden]);
 
-    useMediaSession(video.state, player, onPlayRequested, onPauseRequested, onNextVideoRequested);
+    useMediaSession(video.state, player, onPlayRequested, onPauseRequested, onNextVideoRequested, onPreviousTrackRequested);
 
     React.useEffect(() => {
         const onMediaKey = (action) => {
             switch (action) {
                 case 'play-pause':
-                    video.state.paused ? onPlayRequested() : onPauseRequested();
+                    if (video.state.paused !== null) {
+                        video.state.paused ? onPlayRequested() : onPauseRequested();
+                    }
+                    break;
+                case 'play':
+                    onPlayRequested();
+                    break;
+                case 'pause':
+                    onPauseRequested();
                     break;
                 case 'next-track':
                     if (player.nextVideo !== null) {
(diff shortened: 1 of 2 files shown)


kKaskak (approved):

kKaskak:
thank you

tymmesyde:
Hi, why was the previoustrack action added if it's only rewinding the stream?

kKaskak:
> Hi, why was the previoustrack action added if it's only rewinding the stream?

Technically we don't have a prevEpisode in core so relying on going back in the window is not the best way of handling this 

Most apps first rewind with prev button then on second click go to prev one 

Since we don't have a good way of handling this imo; the implementation only rewinds for now.

erik-kroon:
> > Hi, why was the previoustrack action added if it's only rewinding the stream?
> 
> Technically we don't have a prevEpisode in core so relying on going back in the window is not the best way of handling this
> 
> Most apps first rewind with prev button then on second click go to prev one
> 
> Since we don't have a good way of handling this imo; the implementation only rewinds for now.

That makes sense. My intent was to expose the existing previous-track behavior through Media Session, not to introduce previous-episode navigation.

After thinking about the AirPods/media-key UX though, I agree this may not be ideal for long-form video. Since `previoustrack` only rewinds/restarts the current stream right now, an accidental hardware/media-key previous action could restart an entire episode or movie, which feels more surprising than helpful.

I think the safer follow-up is to remove the Media Session `previoustrack` handler for now, while keeping the existing internal previous-track behavior unchanged.

If rewind-style behavior is desired for hardware/media controls later, `seekbackward` with a small interval seems like a better semantic fit. And if core exposes a proper `prevEpisode`/previous-video action in the future, `previoustrack` can be reintroduced with the expected behavior.

kKaskak:
> > > Hi, why was the previoustrack action added if it's only rewinding the stream?
> 
> > 
> 
> > Technically we don't have a prevEpisode in core so relying on going back in the window is not the best way of handling this
> 
> > 
> 
> > Most apps first rewind with prev button then on second click go to prev one
> 
> > 
> 
> > Since we don't have a good way of handling this imo; the implementation only rewinds for now.
> 
> 
> 
> That makes sense. My intent was to expose the existing previous-track behavior through Media Session, not to introduce previous-episode navigation.
> 
> 
> 
> After thinking about the AirPods/media-key UX though, I agree this may not be ideal for long-form video. Since `previoustrack` only rewinds/restarts the current stream right now, an accidental hardware/media-key previous action could restart an entire episode or movie, which feels more surprising than helpful.
> 
> 
> 
> I think the safer follow-up is to remove the Media Session `previoustrack` handler for now, while keeping the existing internal previous-track behavior unchanged.
> 
> 
> 
> If rewind-style behavior is desired for hardware/media controls later, `seekbackward` with a small interval seems like a better semantic fit. And if core exposes a proper `prevEpisode`/previous-video action in the future, `previoustrack` can be reintroduced with the expected behavior.

Would you like to make a change of removing it then as well? It was introduced before your changes but you could also work on it if you have the ability. Thank you 

For now let's not introduce seeking with the media keys

erik-kroon:
> > > > Hi, why was the previoustrack action added if it's only rewinding the stream?
> > 
> > 
> > > 
> > 
> > 
> > > Technically we don't have a prevEpisode in core so relying on going back in the window is not the best way of handling this
> > 
> > 
> > > 
> > 
> > 
> > > Most apps first rewind with prev button then on second click go to prev one
> > 
> > 
> > > 
> > 
> > 
> > > Since we don't have a good way of handling this imo; the implementation only rewinds for now.
> > 
> > 
> > That makes sense. My intent was to expose the existing previous-track behavior through Media Session, not to introduce previous-episode navigation.
> > After thinking about the AirPods/media-key UX though, I agree this may not be ideal for long-form video. Since `previoustrack` only rewinds/restarts the current stream right now, an accidental hardware/media-key previous action could restart an entire episode or movie, which feels more surprising than helpful.
> > I think the safer follow-up is to remove the Media Session `previoustrack` handler for now, while keeping the existing internal previous-track behavior unchanged.
> > If rewind-style behavior is desired for hardware/media controls later, `seekbackward` with a small interval seems like a better semantic fit. And if core exposes a proper `prevEpisode`/previous-video action in the future, `previoustrack` can be reintroduced with the expected behavior.
> 
> Would you like to make a change of removing it then as well? It was introduced before your changes but you could also work on it if you have the ability. Thank you
> 
> For now let's not introduce seeking with the media keys

#1254

PR #1258, opened 2026-05-04
Dev: Move core logic to a provider

tymmesyde:


src/App/App.js
```diff
@@ -3,44 +3,40 @@
...
-            appVersion: process.env.VERSION,
-            shellVersion: null
-        });
         return {
-            core,
             shell: new Shell(),
             chromecast: new Chromecast(),
             keyboardShortcuts: new KeyboardShortcuts(),
-            dragAndDrop: new DragAndDrop({ core })
+            dragAndDrop: new DragAndDrop({ core: null })
```
kKaskak: is this safe? wouldnt it fail when torrent file is dropped?

kKaskak (changes requested):
you deleted the `encodeStream` wrapper which means `usePlayUrl` will fail?

kKaskak:
https://github.com/Stremio/stremio-web/blob/de2b9d9378ee1711b50df4280c82cb78d3d9777c/src/services/ServicesContext/types.d.ts this is stale?

src/App/App.js
tymmesyde: it was set as draft but thank you

PR #1262, opened 2026-05-06
Player: Improve SideDrawer mobile landscape styles

kKaskak:
videos list was not visible on phone-landscape break point, improved Dropdown styles for this case as well


Botsy (changes requested):
Please, make sure to scroll to top the videos list once the season is changed from dropdown in sidebar. Everything else looks good 👍

Botsy (approved):
tested and looks good 👍

PR #1263, opened 2026-05-06
Player: Correctly respect PWA safe areas

kKaskak:


Botsy (approved):
Tested and looks good 👍

PR #1269, opened 2026-05-12
MetaDetails: Videos list scroll to top on season change

Botsy:
This PR handles the following:

**Season changed scroll:** Scroll the episodes list to the top when a different season is selected.
**Back navigation scroll:** Preserve the episodes list scroll position when navigating back from the streams list.


kKaskak (changes requested):
so there is a bug when you reload the page on details it does not work it, to repro 


 1. Scroll the episodes list down, far enough that scrollTop > 0.
 2. Click an episode so saveScrollPosition runs and you go to the streams/player path.
 3. go back to details page

or just switch season and it wont scroll back down

kKaskak (approved):

PR #1271, opened 2026-05-13
Dev: Move shell logic to provider

tymmesyde:


src/common/Platform/shell/useShell.ts
```diff
@@ -0,0 +1,125 @@
+import { useEffect, useState } from 'react';
+import EventEmitter from 'eventemitter3';
+
+const IPC = globalThis?.chrome?.webview;
+
+const LEGACY_IPC = globalThis.qt.webChannelTransport;
```
Botsy: I get an error in browser that `globalThis.qt` is undefined and it breaks the app. If I change this line to `const LEGACY_IPC = globalThis?.qt?.webChannelTransport || {};` the app loads.

src/routes/Player/Player.js
```diff
@@ -424,7 +417,7 @@ const Player = ({ urlParams, queryParams }) => {
                 seriesInfo: player.seriesInfo,
             }, {
                 chromecastTransport: services.chromecast.active ? services.chromecast.transport : null,
-                shellTransport: services.shell.active ? services.shell.transport : null,
+                shellTransport: platform.shell.active ? platform.shell.transport : null,
```
Botsy: Can you please add `transport` to the `Shell` interface?

src/common/Platform/shell/useShell.ts
```diff
@@ -0,0 +1,125 @@
...
+        windowHidden: false,
+    });
+
+    const on = (name: string, listener: (arg: any) => void) => events.on(name, listener);
+    const off = (name: string, listener: (arg: any) => void) => events.off(name, listener);
+
+    const send = (method: string, ...args: (string | number | object)[]) => {
+        try {
+            IPC?.postMessage(JSON.stringify({
+                id: 0,
```
Botsy: Is the id hardcoded on purpose?

Botsy (commented):
Tested and I couldn't start it locally without the changes I posted in a comment. @tymmesyde can you please check? I believe this is not a solution though.

I tried playing a few streams and the subtitles were disabled for all of them, also they weren't loading even though they are not disabled in settings.

src/common/Platform/shell/useShell.ts
tymmesyde: my bad, should be good now

src/common/Platform/shell/useShell.ts
tymmesyde: Yes, it's a remains of this protocol which serves no purpose anymore

Botsy (approved):

kKaskak (approved):

kKaskak:
regression found: The refactor moved shell IPC/readiness into PlatformProvider /
`src/common/Platform/shell/useShell.ts`, so cold-start open-media can be emitted before App installs its `shell.on('open-media')` route handler.

PR #1273, opened 2026-05-13
feat(addons): reorder installed addon

Aztup:
Adds up/down buttons on each installed addon row to reorder them.
New order is persisted via addonCollectionSet endpoint. Locked addons (like Cinemeta) can't be re ordered.


kKaskak:
hey, i really appreciate the PR but for now we are not allowing reordering addons thank you

PR #1274, opened 2026-05-14
Settings: Add Nvidia GPU Video Processing option

Aztup:


kKaskak:
can you name it enable gpu video processing or something? or is it nvidia only

we merged your shell pr by mistake without bumping the mpv version we need to wait for that to happen and then introduce the setting if its supposed to be one

Aztup:
I believe Amd also has a similar feature, so renaming it to “gpuVideoProcessing” might be a better idea but I don’t know if you can do the same trick with mpv on Amd and I don’t have an Amd gpu to test.

kKaskak:
> I believe Amd also has a similar feature, so renaming it to “gpuVideoProcessing” might be a better idea but I don’t know if you can do the same trick with mpv on Amd and I don’t have an Amd gpu to test.

yeah just rename it for now

kKaskak:
please resolve conflicts

Aztup:
Should be good now

kKaskak:
@Aztup can you update your core version here thx

kKaskak (dismissed):

tymmesyde:
Hi, could you explain what's the logic behind this setting?

Aztup:
> Hi, could you explain what's the logic behind this setting?

To make it simple, it allows Stremio to support the nvidia RTX HDR and VSR on MPV by passing the correct props

kKaskak:
https://github.com/Stremio/stremio-shell-ng/pull/73
https://github.com/Stremio/stremio-video/pull/145

kKaskak (approved):

PR #1277, opened 2026-05-18
Player: retry audio-track restore until player acknowledges selection (fix #2564)

omrihefez:
Fixes #2564 (audio language not applied on stream re-open, LG webOS / Smart TV).

## Bug

On stream re-open, the auto-selection effect in `Player.js` calls `video.setAudioTrack(id)` exactly once when the `audioTracks` list first becomes available, then sets `defaultAudioTrackSelected.current = true` and never retries.

On platforms where the underlying player has not fully initialized at that moment (observed reliably on LG webOS Smart TV), the call is silently ignored. The UI state (`selectedAudioTrackId` in core) reflects the desired track, but actual playback uses the player default. The user observes that the correct language is highlighted in the menu, yet re-clicking it is what actually switches the audio.

## Fix

```js
React.useEffect(() => {
    if (defaultAudioTrackSelected.current) return;
    const savedTrackId = player.streamState?.audioTrack?.id;
    const savedTrack = savedTrackId ? findTrackById(video.state.audioTracks, savedTrackId) : null;
    const audioTrack = savedTrack ?? findTrackByLang(video.state.audioTracks, settings.audioLanguage);
    if (!audioTrack || !audioTrack.id) return;

    if (video.state.selectedAudioTrackId === audioTrack.id) {
        // Player acknowledged the selection — done.
        defaultAudioTrackSelected.current = true;
        return;
    }

    // First request, or the previous one was ignored. Retry.
    video.setAudioTrack(audioTrack.id);
}, [video.state.audioTracks, video.state.selectedAudioTrackId, player.streamState]);
```

Key changes:
- Do not mark the selection as "done" on the first `setAudioTrack` call. Wait until `video.state.selectedAudioTrackId` actually matches the desired id, which is the signal that the player has applied it.
- Include `video.state.selectedAudioTrackId` in the effect deps so we re-run when the player updates its state.

`setAudioTrack` is idempotent on the core side, so re-issuing the same id is safe.

## Testing

I do not have a webOS dev kit handy. The bug reporter (#2564) reproduces every time on LG C5 OLED. The patch is minimal and the new behavior gracefully degrades to the old behavior on platforms where the first call already takes (the equality check short-circuits the retry path immediately).

Happy to iterate if maintainers see a cleaner approach — e.g., a max-retry cap, or moving this into core so non-web players benefit too.


kKaskak:
i dont know what you did there but something went wrong, also this is not the lg tv app

omrihefez:
Thanks for taking a look, @kKaskak — and apologies for the confusion. Two clarifications:

1. **Intent of the patch**: I targeted `src/routes/Player/Player.js` because that's where the audio-track restore effect runs on stream open, and I see the issue (audio track persists in UI state but the underlying player initializes with the default until re-selected) reproduce on the web build as well as the LG smart-TV channel. I assumed the smart-TV channel was built from this repo; sounds like that's wrong.

2. **Where should I file the fix instead?** If the LG TV app lives in a non-public repo, I'm happy to keep the bug report (Stremio/stremio-bugs#2564) as the visible record and drop this PR. If it does live somewhere I can contribute to, point me at it and I'll port the patch.

CLA: I can sign it if useful — let me know if there's a path forward and I'll do that first.

Happy to take this offline if there's a better channel.

PR #1286, opened 2026-05-21
feat: add dual subtitle support for language learning (desktop only)

lakshayrastogi:
## Summary

Adds a "Secondary Subtitle" section to the player subtitles menu, allowing users to select a second subtitle track that renders simultaneously with the primary one. Gated to desktop only (`platform.shell.active`) since it relies on mpv's `secondary-sid`. Primary subtitle selection behavior is completely unchanged.

### Motivation
Enables language learning workflows - e.g., watching French content with French as primary and English as secondary subtitles simultaneously.

### Changes

**`src/routes/Player/useVideo.js`**
- Added `selectedSecondarySubtitlesTrackId` to video state
- Added `setSecondarySubtitlesTrack(id)` function (does NOT nullify primary)
- Exported from hook

**`src/routes/Player/useSubtitles.ts`**
- Added `selectSecondaryTrack` callback
- `disableSubtitles()` also clears secondary track
- Added `isShellActive` and secondary track info to `menuProps`
- Updated `toggleSubtitles` shortcut to include secondary track state
- Updated all dependency arrays

**`src/routes/Player/SubtitlesMenu/SubtitlesMenu.js`**
- Added secondary subtitle section (renders only when: primary is selected + `isShellActive` + callback exists)
- Separate language list for secondary selection (excludes primary language)
- Click handlers: select a language or click "Off" to disable
- Secondary indicator: hollow circle icon vs filled circle for primary
- "Secondary Subtitle" header with separator line
- All existing click handlers preserved (no interference)

**`src/routes/Player/SubtitlesMenu/styles.less`**
- Added `.secondary-header`, `.secondary-selected`, `.secondary-icon` styles

**`src/routes/Player/SubtitlesMenu/SubtitleVariant/SubtitleVariant.less`**
- Added `.secondary-icon` style only (variant component itself is unchanged)

**`src/routes/Player/useSubtitles.d.ts`**
- Updated type definitions for new fields

### Platform Gating
The secondary section only appears when:
1. A primary subtitle is selected
2. Running in desktop shell (`platform.shell.active` - established pattern)
3. Callback is provided

Invisible on web, mobile, and TV platforms.

### Translation
`PLAYER_SUBTITLES_SECONDARY` key used with `defaultValue: 'Secondary Subtitle'` fallback.

### Related
- stremio-video: adds mpv secondary-sid bridge
- stremio-translations: adds translation key

### Related PRs
- stremio-video: https://github.com/Stremio/stremio-video/pull/146
- stremio-translations: https://github.com/Stremio/stremio-translations/pull/1053

### DIFF
--- a/src/routes/Player/SubtitlesMenu/SubtitleVariant/SubtitleVariant.less
+++ b/src/routes/Player/SubtitlesMenu/SubtitleVariant/SubtitleVariant.less
@@ -48,6 +48,16 @@
         margin-left: 1rem;
         background-color: var(--secondary-accent-color);
     }
+
+    .secondary-icon {
+        flex: none;
+        width: 0.5rem;
+        height: 0.5rem;
+        border-radius: 100%;
+        margin-left: 1rem;
+        border: 2px solid var(--secondary-accent-color);
+        background-color: transparent;
+    }
 }
 
 .context-menu-option {
--- a/src/routes/Player/SubtitlesMenu/SubtitlesMenu.js
+++ b/src/routes/Player/SubtitlesMenu/SubtitlesMenu.js
@@ -75,6 +75,18 @@ const SubtitlesMenu = React.memo(React.forwardRef((props, ref) => {
                 :
                 null;
     }, [subtitlesTracks, extraSubtitlesTracks, props.selectedSubtitlesTrackId, props.selectedExtraSubtitlesTrackId]);
+    const selectedSecondarySubtitlesLanguage = React.useMemo(() => {
+        return typeof props.selectedSecondarySubtitlesTrackId === 'string' ?
+            allSubtitles
+                .reduce((selected, { id, lang }) => {
+                    if (id === props.selectedSecondarySubtitlesTrackId) {
+                        return lang;
+                    }
+                    return selected;
+                }, null)
+            :
+            null;
+    }, [allSubtitles, props.selectedSecondarySubtitlesTrackId]);
     const subtitlesTracksForLanguage = React.useMemo(() => {
         const tracks = allSubtitles.filter(({ lang }) => lang === selectedSubtitlesLanguage);
         return sortByValues(tracks, ORIGIN_PRIORITIES);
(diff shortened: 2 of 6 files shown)


kKaskak:
how can you learn a language with double subs? :D I cannot comprehend why would somebody want to read in a non native lang when learning it. this is super hard to do, the subs are usually so fast that for the person who learns a language its the opposite, you actually choose a non native audio and native subs to get the hear and understand that lang you are learning, for most people its impossible to read so fast in the lang you are learning. im sorry but i feel like this feature is pointless and it adds uneccessary complexity to the Player and the app itself.

lakshayrastogi:
Well, I have found that as an intermediate learner of French now, I can
watch movies with french audio and english subtitles but i need to keep
pausing and checking the words because I don't yet know all the words.
Usually I will pause, repeat the audio and possibly look up the french word
for the english word (they don't sound like they are written many times
which makes this harder too!) - this is tedious to do and takes away from
the joy of watching. Think of the dual subtitles as being training wheels
because I get to see the translation right there after at max pausing for a
split second if at all. I actually hooked this up in an external player
through Stremio and found it helpful but I have to jump through a few hoops
to do it on every stream that I need to. Plus what I would really like to
do is add this on the tv app because I haven't found a way to do that
natively and I don't want to cast from my laptop every time - I know this
set of PRs is for the desktop but I was hoping I could add it as a next set
once this went through.

Lastly, secondary subtitles are built into many libraries in general so
regardless of my motivation to use it for language learning, it's an
accepted norm.

Appreciate your response either way and on a separate note, thank you so
much for building stremio - I have consistently evangelised it amongst my
friends (probably in over a 100 households) wherever I have gone and
everyone loves it. I love it too!

mrcanelas:
I think this could be easily implemented as an add-on, which would work on all platforms.

lakshayrastogi:
> I think this could be easily implemented as an add-on, which would work on all platforms.

@mrcanelas thanks for the suggestion, I tried this - the problem is that finding the two subtitles that work together from the lists returned by opensubtitles is a hard problem and it really needs a user to compare and decide if the timing and lyrics match. For now I have made the addon return all combinations (little clunky but will do for now) but the ideal solution would give controls to the user in the player to select a second language from the same list of subtitles as the first.

mrcanelas:
> > I think this could be easily implemented as an add-on, which would work on all platforms.
> 
> 
> 
> @mrcanelas thanks for the suggestion, I tried this - the problem is that finding the two subtitles that work together from the lists returned by opensubtitles is a hard problem and it really needs a user to compare and decide if the timing and lyrics match. For now I have made the addon return all combinations (little clunky but will do for now) but the ideal solution would give controls to the user in the player to select a second language from the same list of subtitles as the first.

Can you send the code? Did you compare the subtitle names? Ideally, you should use the existing OpenSubtitle add-on and combine the subtitles it returns for each IMDb ID. You can also analyze the timing of the dialogues within the subtitles to synchronize them.

Pigamer37:
> I think this could be easily implemented as an add-on, which would work on all platforms.

It already exists! It's called strelingo

kKaskak:
> > I think this could be easily implemented as an add-on, which would work on all platforms.
> 
> @mrcanelas thanks for the suggestion, I tried this - the problem is that finding the two subtitles that work together from the lists returned by opensubtitles is a hard problem and it really needs a user to compare and decide if the timing and lyrics match. For now I have made the addon return all combinations (little clunky but will do for now) but the ideal solution would give controls to the user in the player to select a second language from the same list of subtitles as the first.

https://github.com/Serkali-sudo/strelingo-addon/blob/main/README.md i guess this addon does just that? would this solve your problem?

kKaskak:
closing due to addon being the solution

PR #1289, opened 2026-05-25
fix: handle malformed route params without crashing

### LINKED ISSUE #1288: [Bug]: Malformed encoded route params crash the router
### Stremio-Web Version

/development branch

### Browser

Chrome

### Platform / Device type

Linux

### What Happened?

Opening a route whose path parameter contains malformed percent encoding crashes route parsing before the app can render a route or fallback. For example, `#/metadetails/movie/%E0%A4%A` matches the meta details route, then `urlParamsForPath` calls `decodeURIComponent('%E0%A4%A')`, which throws `URIError: URI malformed`.

Expected behavior: malformed external or manually entered links should not crash the router. The app should keep rendering and pass through the raw parameter when it cannot be decoded.

Actual behavior: the exception escapes during route parameter parsing and breaks navigation/rendering.

### Logs

```shell
URIError: URI malformed
    at decodeURIComponent (<anonymous>)
    at src/router/Router/urlParamsForPath.js
```

### Notes

This is reproducible directly from the current `development` branch route parser with:

```shell
node -e "const routes=require('./src/common/routesRegexp'); const urlParams=require('./src/router/Router/urlParamsForPath'); console.log(urlParams(routes.metadetails, '/metadetails/movie/%E0%A4%A'));"
```

### Code of Conduct

- [x] I agree

officialasishkumar:
Fixes #1288.

## What changed
- Added a safe route-parameter decoder that preserves malformed percent-encoded values instead of throwing.
- Updated router parameter parsing and the meta details episode picker to use the safe decoder.
- Added Jest coverage for normal decoded route params and malformed route params.

## Why
Malformed external or manually entered hash routes such as `#/metadetails/movie/%E0%A4%A` currently throw `URIError: URI malformed` during route parsing, which can break navigation before the app can render a route or fallback.

## Validation
- `node -e "const routes=require('./src/common/routesRegexp'); const urlParams=require('./src/router/Router/urlParamsForPath'); console.log(JSON.stringify(urlParams(routes.metadetails, '/metadetails/movie/%E0%A4%A')));"`
- `pnpm jest tests/urlParamsForPath.spec.js --runInBand`
- `pnpm lint`
- `pnpm build`
- `pnpm jest --runInBand`

### DIFF
--- a/src/common/safeDecodeURIComponent.js
+++ b/src/common/safeDecodeURIComponent.js
@@ -1,0 +1,11 @@
+// Copyright (C) 2017-2026 Smart code 203358507
+
+const safeDecodeURIComponent = (value) => {
+    try {
+        return decodeURIComponent(value);
+    } catch (_) {
+        return value;
+    }
+};
+
+module.exports = safeDecodeURIComponent;
--- a/src/router/Router/urlParamsForPath.js
+++ b/src/router/Router/urlParamsForPath.js
@@ -1,10 +1,12 @@
 // Copyright (C) 2017-2023 Smart code 203358507
 
+const safeDecodeURIComponent = require('../../common/safeDecodeURIComponent');
+
 const urlParamsForPath = (routeConfig, path) => {
     const matches = path.match(routeConfig.regexp);
     return routeConfig.urlParamsNames.reduce((urlParams, name, index) => {
         if (Array.isArray(matches) && typeof matches[index + 1] === 'string') {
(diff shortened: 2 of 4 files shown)


kKaskak:
does this happen with the react router PR? if not im closing this since we are merging it soon

kKaskak:
?

kKaskak:
closing due inactivity

PR #1290, opened 2026-05-26
Player: Remaining time accounts for playback speed

Aztup:
The remaining time mode now supports playback speed `remainingDuration / playbackSpeed`


Botsy:
I just tested on the development branch and on my side the remaining time seems to be updating correctly without those changes.

Aztup:
If we take MPV for example, when playing at 2x on a 40 minutes video it’ll display 20 min remaining

Botsy (approved):
Looks good to me.

PR #1291, opened 2026-05-26
Player: Redesign Speed Menu

Aztup:


kKaskak:
would you agree making it a slider would solve the issue between 0.25x and 8x

Aztup:
Yes that would fix the issue too.

I was also thinking about a slider too but it’s less precise and would use more space.

A slider is fine, we could make the ui smaller in height and make it wider instead to have more slider control

The values would be 
5(Max), 0.1(Min), 0.1(Step)

kKaskak:
> Yes that would fix the issue too.
> 
> I was also thinking about a slider too but it’s less precise and would use more space.
> 
> A slider is fine, we could make the ui smaller in height and make it wider instead to have more slider control
> 
> The values would be 5(Max), 0.1(Min), 0.1(Step)

slider inside the menu and no options at all? wdyt

Aztup:
Maybe with preset options at the bottom?
Something like that
image

kKaskak:
> Maybe with preset options at the bottom? Something like that image

yeah make it use the inbuilt slider component that we have 
also top picks at the bottom , use defaults from vlc if you can

Aztup:
Yeah I saw the inbuilt slider component and planned to use it, this was just a design idea.

> also top picks at the bottom , use defaults from vlc if you can

I don't think vlc has preset options

image

Aztup:
I kept the default presets Stremio had already.
This is the new version
image

Botsy (dismissed):
Tested and looks good to me 👍

Botsy (dismissed):

kKaskak (dismissed):

kKaskak (approved):

Botsy (approved):

PR #1294, opened 2026-06-01
Allow direct subtitle delay entry

Tavaresiqueira:
## Summary

This makes the subtitle delay value in the player subtitles menu directly editable while keeping the existing +/- buttons at 0.25s increments.

The motivation is a real playback case where subtitles were more than 40 seconds out of sync. With only 0.25s steps, reaching the rough offset requires too many repeated clicks before fine tuning can even start.

## Behavior

- Users can type the delay in seconds directly in the existing delay stepper.
- Enter or blur commits the typed value; Escape restores the current value.
- The +/- buttons still adjust by 0.25s for precise tuning after the rough offset is entered.
- Size and vertical position steppers remain display-only values with buttons, matching their previous behavior.
- The existing `subtitleDelay` stream state path is unchanged, so remembered subtitle timing continues to work as before.

## Validation

- `npx --yes pnpm@10.12.1 lint`
- `npx --yes pnpm@10.12.1 build`

The production build still reports the existing webpack asset/entrypoint size warnings.

### DIFF
--- a/src/routes/Player/SubtitlesMenu/Stepper/Stepper.less
+++ b/src/routes/Player/SubtitlesMenu/Stepper/Stepper.less
@@ -38,11 +38,44 @@
             }
         }
 
-        .value {
+        .value, .editable-value {
             flex: 1;
             font-weight: 500;
             text-align: center;
             color: var(--primary-foreground-color);
+        }
+
+        .editable-value {
+            display: flex;
+            flex-direction: row;
+            align-items: center;
+            justify-content: center;
+            min-width: 0;
+            user-select: text;
+
+            .input {
+                flex: 0 1 5rem;
+                min-width: 0;
+                padding: 0;
+                border: none;
+                outline: none;
+                appearance: textfield;
+                font: inherit;
+                color: var(--primary-foreground-color);
+                text-align: right;
+                background: transparent;
+
+                &::-webkit-outer-spin-button,
+                &::-webkit-inner-spin-button {
+                    -webkit-appearance: none;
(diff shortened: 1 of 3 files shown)


kKaskak:
you can hold the button to fast forward to the value you want tho?

kKaskak:
> The motivation is a real playback case where subtitles were more than 40 seconds out of sync. With only 0.25s steps, reaching the rough offset requires too many repeated clicks before fine tuning can even start.

this not true explained in the comment before

joao-siqueira-genoa:
> > The motivation is a real playback case where subtitles were more than 40 seconds out of sync. With only 0.25s steps, reaching the rough offset requires too many repeated clicks before fine tuning can even start.
> 
> this not true explained in the comment before

You're right, the PR description is too strong there. Holding the button means the user does not need repeated individual clicks, so I should not phrase it as only 0.25s manual steps.

The gap I wanted to solve is narrower: when the user already knows the delay is very large, e.g. around 40s, direct entry is still faster and more precise than holding and trying not to overshoot. The existing hold-to-repeat and 0.25s +/- controls remain useful for fine tuning.

I can update the PR description to make that distinction clear, or close the PR if you think hold-to-repeat is enough for this UX.

kKaskak:
for now i think what we have is enough, we do not want to complicate the menus with additional features. also its rare to see that you need -60/120sec of sub delay. even if you exit the stream, the stream bucket should save that offset for you, which means on enter the offset should be applied again without issues

PR #1295, opened 2026-06-01
chore: update dependencies

Rohan1572:
> Updated packages

image
image
image


tymmesyde:
Hi, thank you but we handle the dependencies update ourselves

PR #1300, opened 2026-06-08
fix(stremio-web): stabilize hosted streaming server

WizzyDizzySWE:
## Summary
- route hosted Stremio Web/Core localhost streaming-server calls to the Wallis hosted streaming server
- use an object-based stremio-core-web worker initialization path
- add cache busting for dirty hosted builds and modern PWA metadata
- declare pnpm 11.5.2 for reproducible Docker builds

## Verification
- `pnpm i --frozen-lockfile` passes with pnpm 11.5.2
- Docker image builds successfully
- deployed stremio-web container starts and serves the timestamped worker bundle

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,7 @@ ENV PNPM_HOME="/pnpm"
 ENV PATH="$PNPM_HOME:$PATH"
 
 RUN corepack enable
+RUN corepack prepare pnpm@11.5.2 --activate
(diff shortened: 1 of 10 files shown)


WizzyDizzySWE:
Closing this PR because the patch is Wallis-specific and should live in the WizzyDizzySWE fork rather than upstream Stremio.

PR #1316, opened 2026-06-23
Fix virtual keyboard shortcuts

kKaskak:


Botsy (commented):
Reviewed the changes and pushed a couple of commits to this branch - hope that's ok, easier than writing it all up.

What I pushed:

- A guard for `code === 'Unidentified'` in both helpers. Some browsers (Android WebViews) report the literal string 'Unidentified' instead of an empty code, and since that's truthy, `event.code || ...` never fell back to key/keyCode. So the Player space/seek handlers and fullscreen Escape were still broken on exactly the devices this PR targets. The shortcuts matcher survived it only by accident (the normalized key was also in the array).

- Unit tests for keyboard.ts covering the device profiles (real code, key-only, keyCode-only, code 'Unidentified'), plus a small jest config + @babel/preset-typescript so jest can run TS. The Unidentified test was red before the guard fix and green after, so this stays pinned down.

I verified the key-only cases with synthetic key events in the console as well, but I didn't get to test on an actual device - that's still needed before merge. What did you repro this on originally?

A few other things I noticed along the way:

- Popup.js:41 and ModalDialog.js:45 still do a raw `event.code === 'Escape'`, so closing popups/modals will still be broken on these devices. Fine to leave for a follow-up, just flagging it.

- Some virtual keyboards never set `repeat` on synthesized keydown events, so I'm not sure how the long-press logic behaves there. Worth covering when someone tests on real hardware.

- Small footgun: `getKeyboardShortcutKey` returns the physical code when it's available ('Digit1', 'Equal', ...) while `getKeyboardShortcutKeys` also returns key-style names. It works right now only because Escape/F11/Space/Arrow* are the same as code and key. But if someone later writes `getKeyboardShortcutKey(e) === '='` it will silently fail on physical keyboards. Maybe just add a comment that the singular is only safe for named keys? The only place that actually needs the physical code is the throttle map in Shortcuts.tsx:45, everywhere else could use the normalized key.

Botsy (approved):

PR #1319, opened 2026-06-25
Player: Handle Infuse external player progress

kKaskak:


kKaskak:
paused because of play external removal

src/App/DeepLinkHandler.js
```diff
@@ -14,11 +15,12 @@ const DeepLinkHandler = () => {
...
                 const [, deepLinks] = content;
                 if (typeof deepLinks.metaDetailsVideos === 'string') {
-                    navigate(toPath(deepLinks.metaDetailsVideos));
+                    const path = toPath(deepLinks.metaDetailsVideos);
+                    navigateToRoute(navigate, location, path);
                 }
             }
         }
-    }, [streamingServer.torrent]);
+    }, [streamingServer.torrent, location.pathname, location.search, navigate]);
```
Botsy, line 23: Adding `location.pathname` / `location.search` here changes this from
"navigate once when a torrent becomes Ready" to "re-check on every route change".

Nothing clears `streamingServer.torrent`, and this component is mounted app-wide,
so once a magnet resolves, the effect keeps firing. `navigateToRoute` only skips
when you're already on the torrent's page — which is false precisely when the
user has navigated away. Result: after opening a magnet, clicking a stream to
reach `/player/...` possibly bounces straight back to `/detail/...`. Settings and Board
are equally unreachable.

Could we keep the location out of the deps and read it from a ref, plus a
one-shot guard so a given torrent only navigates once?

PR #1320, opened 2026-06-25
Details: stop addon dropdown scroll jumping

AndnixSH:
The streams addon dropdown reset its scroll back to the selected option every time the option list grew (i.e. each addon finishing loading), because the scroll-into-view effect depended on the `selectedOption` object, which is recreated on every render. Key it on the stable selected `value` instead, so it only runs when the menu opens or the selection actually changes.

~~Also rework the series streams header: move the addon dropdown onto its own full-width row below the back-arrow + episode title (instead of cramming all three into one row), and raise the open menu's max-height to 60vh so more addons are visible at once.~~

### DIFF
--- a/src/components/MultiselectMenu/Dropdown/Dropdown.tsx
+++ b/src/components/MultiselectMenu/Dropdown/Dropdown.tsx
@@ -21,6 +21,7 @@ const Dropdown = ({ level, setLevel, options, onSelect, value, menuOpen }: Props
     const { t } = useTranslation();
     const optionsRef = useRef(new Map());
     const containerRef = useRef(null);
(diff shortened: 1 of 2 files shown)


AndnixSH:
Before:

image

After:

image

kKaskak:
hey could you keep only the re-rendering change and remove the UI changes? if you move out the selector it will cause layout shifts when it shows. please revert the design changes

AndnixSH:
> hey could you keep only the re-rendering change and remove the UI changes? if you move out the selector it will cause layout shifts when it shows. please revert the design changes

I reverted it

kKaskak (approved):

PR #1328, opened 2026-07-07
feat: Update build system to rsbuild

domantasjurkus:
tldr:

- webpack is old and slow, rsbuild is a nice drop-in replacement. The goal of this PR is to investigate the feasibility and effort required to switch the build system.
- From personal experience, switching from webpack to rsbuild on a company project sped up builds from 40 seconds to about 5 seconds.


kKaskak:
hey, what does this pr solve exactly?

domantasjurkus:
@kKaskak Main goal was to speed up the build, however, I see that webpack compiles in <10s with an M1 mac, (~4s with rsbuild), so the benefit may be marginal at this point (plus, thorough QA would be needed for prod).

I'm dabbling a bit with rsbuild/claude from time to time, hence the draft MR. If you prefer, I can close this as this is a POC.

kKaskak:
POC closed

PR #1329, opened 2026-07-07
Auto apply web UI updates

kKaskak:
Adds the web/PWA update flow for web.stremio.com when a waiting service worker is found on open or resume.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,8 @@
         "spatial-navigation-polyfill": "github:Stremio/spatial-navigation#64871b1422466f5f45d24ebc8bbd315b2ebab6a6",
         "stremio-translations": "github:Stremio/stremio-translations#0f290a5e5d17c295083ab0162e0b9054c07b123b",
         "url": "0.11.4",
-        "use-long-press": "^3.3.0"
+        "use-long-press": "^3.3.0",
+        "workbox-window": "7.4.1"
     },
     "devDependencies": {
         "@babel/core": "7.29.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -104,6 +104,9 @@ importers:
       use-long-press:
         specifier: ^3.3.0
         version: 3.3.0(react@18.3.1)
+      workbox-window:
+        specifier: 7.4.1
+        version: 7.4.1
     devDependencies:
       '@babel/core':
         specifier: 7.29.0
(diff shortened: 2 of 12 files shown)


Botsy (dismissed):
## Review — Auto apply web UI updates (#1329)

Tested locally against a production build (`webpack serve --mode production`): the update-on-open and update-on-resume paths both work — `WebUpdateScreen` renders, the 2s progress bar runs, and the `SKIP_WAITING` → activate → `controllerchange` → reload cycle completes. Verified Workbox 7.4.1 auto-generates the matching `SKIP_WAITING` message listener with `skipWaiting: false`, so the apply mechanism is sound. Core React logic (effect deps, cleanup, player-route guard, timing) checks out. Findings below are all minor — nothing blocking.

### 1. Document the "reload on next resume" behavior for focused tabs
`src/common/useServiceWorkerUpdater.ts:151-153`

On an unprompted `controllerchange` (another tab applied the update → `clientsClaim` claims this tab too), the handler sets `reloadPending = true` but `promptReady = false`, so a **focused, already-visible** tab won't reload until its next `visibilitychange → visible`. This is fine as a design choice — a backgrounded tab self-heals on resume before the user can interact with it. The only exposed edge is two windows visible side-by-side where the user keeps using the stale one. Not worth extra code; just add a one-line comment so `promptReady = false` doesn't read as a bug to the next maintainer.

### 2. `applyingUpdate` is set but never reset
`src/common/useServiceWorkerUpdater.ts:130`

Benign in the happy path (the reload wipes module state), but if `SKIP_WAITING` is a no-op (worker already redundant / activation fails), no `controllerchange` follows and `applyingUpdate` stays `true`. A later unrelated `controllerchange` then takes the immediate-`reloadPage()` branch. Suggest clearing the flag after the paired `controllerchange`.

### 3. `reloading` is a permanent one-shot latch with no reset
`src/common/useServiceWorkerUpdater.ts:71-78`

If a `window.location.reload()` is ever suppressed/cancelled (e.g. a `beforeunload` the user dismisses), no further reload can succeed for the page's lifetime — including the legitimate `controllerchange` reload at line 147. Narrow, but worth noting.

### 4. Progress duration is a magic constant split across two files
`src/App/WebUpdateScreen/WebUpdateScreen.tsx:9` and `src/App/WebUpdateScreen/WebUpdateScreen.less:101`

`APPLY_DELAY = 2000` and `animation: 2s ... progress-fill` must stay in sync for the bar to reach 100% exactly when the update applies, but nothing links them. Suggest a shared CSS custom property (or an inline `animationDuration` derived from `APPLY_DELAY`), or at minimum a comment.

---

Overall: solid PR, safe to merge. #2 and #3 are cheap hardening; #1 and #4 are just comments.

Botsy (approved):

PR #1330, opened 2026-07-10
Fix card navigation targets

kKaskak:
- Makes card navigation context-aware for Library and Continue Watching.
- Aligns card and “Details” menu destinations.
- Keeps the Continue Watching play overlay available only when player deeplink is present
- Prevents play and dismiss clicks from also triggering card navigation.
- Removes direct-player fallback from whole-card clicks.


Botsy (approved):
Tested and all looks good 👍

PR #1331, opened 2026-07-10
Separate episode navigation and playback

kKaskak:
adds a play button to Video component with direct playback or youtube videos


Botsy (approved):
lgtm 👍

PR #1348, opened 2026-07-21
Player: Keep subtitles above controls

kKaskak:
Measures the visible player controls and applies them as the temporary minimum from Stremio/stremio-video#159.


Botsy (dismissed):
lgtm

Botsy (approved):

PR #1349, opened 2026-07-22
Fix repeated context menu handling

kKaskak:
Prevents open WebUI context menus from falling through to native browser menus. Keeps Shift+right-click available for inspection with Stremio/stremio-shell-ng#87.


TRtomasz (approved):
tested on windows
Both context menus are now working fine

PR #1360, opened 2026-08-01
Player: Prevent subtitle selection from context menu

kKaskak:
Stop portal context-menu clicks from bubbling into the subtitle row while keeping menu actions and auto-close unchanged.


Botsy (approved):
tested and lgtm 👍

PR #1363, opened 2026-08-03
Wait before applying web UI updates

kKaskak:


Botsy (dismissed):

Botsy (approved):

Botsy (dismissed):

kKaskak:
```
an update during a first-visit session can get stuck in “Try again.”
     The activation handler
     (https://github.com/Stremio/stremio-web/blob/25d1c72a357d630bf020fe9e0d2528b5e15130cb/src/App/WebUpdateScreen/useServiceWorkerUpdater.ts#L99-L109) ignores
     events where isUpdate is false. Workbox captures that flag when registration starts; it stays false if the page initially had no worker.
     ---
the two-release rollout doesn’t cover clients that skip phase one.
     With skipWaiting: false (https://github.com/Stremio/stremio-web/blob/d362f27fe8b62fac2cfdc4cc13d56b7c920b6a13/webpack.config.js#L223-L228), a client that
     never received #1329 can keep loading its old cached UI, which has no handler to activate the waiting worker.
```

src/App/WebUpdateScreen/useServiceWorkerUpdater.ts
```diff
@@ -6,14 +6,13 @@ import { usePlatform } from 'stremio/common';
...
-const APPLY_TIMEOUT = 15 * 1000;
+const APPLY_TIMEOUT = 3 * 1000;
 
 type UpdaterState =
     | { status: 'idle' }
     | { status: 'ready', autoApply: boolean }
     | { status: 'applying' }
-    | { status: 'reload-ready' }
-    | { status: 'failed' };
+    | { status: 'reload-ready' };
```
Botsy, line 28: Dropping the `failed` state removes the only bail-out here. If `skipWaiting`
never takes effect, `waiting` keeps firing with `wasWaitingBeforeRegister`
true, so `autoApply` stays true and we reload every ~3s with no way out.
I haven't reproduced this - with `clientsClaim: true` the normal path should
work - but is it worth keeping a failure state after N attempts, just in case?

Botsy (approved):

PR #1368, opened 2026-08-03
fix: honor external player selection for library items

defenj:
- Makes the Library and Continue Watching Play actions honor `deepLinks.externalPlayer`, with the same link precedence as the stream rows (web, then `openPlayer` for the platform, then the playlist, then the internal player)
- Covers the play overlay, the Play menu action, and Continue Watching through `ContinueWatchingItem`
- External players report no progress back, so resume behavior is unchanged (Stremio/stremio-bugs#1824)


kKaskak:
hello external player feature is disabled right now for both windows and macos so this addition wont make it to the app as of now. im sorry

PR #1370, opened 2026-08-04
Player: Preserve subtitle language while binge watching

kKaskak:
Stacked on #1367.
Depends on Stremio/stremio-core#1023.
Related to Stremio/stremio-bugs#2701.

---
X = global setting and Y = player-session choice.

No active session and no saved track:

  - X available embedded only → embedded X.
  - X available external only → external X.
  - X available in both → embedded X.
  - X unavailable, English available → controller unchanged; English is not selected.
  - X and English unavailable → controller unchanged.
  - Controller already selected something → selection remains unchanged.
  - X appears later → X is selected.
  - Global setting Off → subtitles explicitly off.

  Saved per-stream track:

  - Exact saved track available → exact track restored.
  - Exact track missing, saved Y available → another Y track.
  - Saved Y available in both → saved source preferred.
  - Valid saved Y restored → session Y is seeded.
  - Saved Y unavailable after seeding → global X.
  - Saved Y and X unavailable after seeding → English.
  - Unknown saved language, exact track available → exact track only.
  - Unknown saved language, exact track missing → global X only; no English fallback.
  - Global setting Off → saved track ignored; subtitles off.
  - Active session Y versus saved X → session Y wins.
  - Active session Y plus saved exact Y → exact Y preferred when its source matches.

  Active session Y:

  - External Y selected, external Y available → external Y.
  - External Y selected, only embedded Y available → embedded Y.
  - External Y available in both → external Y.
  - Embedded Y selected, embedded Y available → embedded Y.
  - Embedded Y selected, only external Y available → external Y.
  - Embedded Y available in both → embedded Y.
  - Y unavailable, X available → X, preferred source first.
  - Y and X unavailable, English available → English, preferred source first.
  - Y, X and English unavailable → subtitles off.
  - Global setting Off, Y available → Y.
  - Global setting Off, Y unavailable → English.
  - No recognized Y → global X, then English.
  - Y appears later → automatically switches back to Y.
  - Temporary X fallback → Y remains remembered.
  - Temporary English fallback → Y remains remembered.
  - User manually selects X → session preference becomes X.
  - User switches source for Y → new source becomes preferred.
  - Multiple Y tracks → current or exact variant retained.
  - Y track disappears → next matching Y track used.
  - Track results reorder → current equal-priority track retained.
  - No tracks initially → waits for matching tracks.
  - No arbitrary first-track fallback → only Y, X or English.

  Disabled and lifecycle cases:

  - User disables subtitles → following episodes remain off.
  - Disabled session with saved track → remains off.
  - User enables subtitles again → restores Y fallback chain.
  - Disabling subtitles → clears the current exact saved track.
  - Player remains open → session preference survives episode loads.
  - Player exits → session preference clears.
  - New player session → saved/global behavior applies again.
  - Global X changes during session → Y remains first; new X becomes fallback.
  - Automatic fallback selection → does not overwrite Y or the stream bucket.
  - Manual selection → updates session Y and exact stream track.
  - Equivalent language codes → treated as the same language.
  - Unknown language code → cannot transfer reliably between episodes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
         "@babel/runtime": "7.29.2",
         "@sentry/browser": "8.42.0",
         "@stremio/stremio-colors": "5.2.0",
-        "@stremio/stremio-core-web": "https://stremio.github.io/stremio-core/stremio-core-web/feat/player-subtitle-preference/stremio-stremio-core-web-0.60.2.tgz",
+        "@stremio/stremio-core-web": "https://stremio.github.io/stremio-core/stremio-core-web/feat/player-subtitle-language-preference/stremio-stremio-core-web-0.60.2.tgz",
         "@stremio/stremio-icons": "5.10.0",
         "@stremio/stremio-video": "0.0.90",
         "a-color-picker": "1.2.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -18,8 +18,8 @@ importers:
         specifier: 5.2.0
         version: 5.2.0
       '@stremio/stremio-core-web':
-        specifier: https://stremio.github.io/stremio-core/stremio-core-web/feat/player-subtitle-preference/stremio-stremio-core-web-0.60.2.tgz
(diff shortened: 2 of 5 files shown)


Botsy (approved):
Tested and all works as expected 👍

PR #1376, opened 2026-08-16
fix(player): stop local playback, send subtitles and stop the device when casting

dioKaratzas:
### Problem

Desktop casting goes through the streaming server (`StreamingServer` / `PlayOnDevice`), and `stremio-web` only sends that one action, then waits for the `PlayingOnDevice` core event. Three bugs follow from that:

- Stremio/stremio-bugs#2731 — local playback keeps running until the event arrives, so the stream plays locally *and* on the device.
- Stremio/stremio-bugs#2732 — the selected subtitles track is never sent to the device, so subtitles render locally only.
- Stremio/stremio-bugs#2733 — leaving the player never stops the device, which keeps playing with no way to control it from the app.

The server already speaks the stremio-cast protocol on `/casting/{deviceId}/player`: a `POST` sets player properties and returns the full state, `subtitlesSrc` carries the subtitles URL, and `source: null` stops playback. Stremio 4.4 used exactly this through the `stremio-cast` client; v5 stopped using everything except the initial play.

Verified against a running streaming server (v4 shell, `GET /casting/{id}/player` on a live session):

```json
{"volume":1,"time":406192.1,"paused":false,"state":3,"length":3162460,
 "source":"http://127.0.0.1:11470/<hash>/0",
 "subtitlesSrc":"https://subs5.strem.io/.../file/...","subtitlesDelay":0,"subtitlesSize":2}
```

### Changes

- New `useCastDevice` hook wrapping the `/casting/{deviceId}/player` endpoint (set subtitles, stop).
- `onCastDeviceSelected` pauses local playback immediately instead of waiting for `PlayingOnDevice`.
- The selected extra subtitles track is pushed on cast start and whenever it changes while casting.
- The device is stopped (`source: null`) when the player unmounts.

### Notes

- Only `subtitlesSrc` is sent. `subtitlesDelay` / `subtitlesSize` are supported by the endpoint too, but their units/scale differ from the in-app settings, so they are left alone here — happy to add them if you can confirm the mapping.
- An alternative shape would be a `StopOnDevice` / `SetDeviceSubtitles` action in `stremio-core` next to `PlayOnDevice`, keeping all server calls in core. Say the word and I'll move it there instead.

### Testing

- `eslint` clean, `tsc --noEmit` shows no new errors, production build passes.
- Endpoint behavior verified against a live streaming server casting session.

### DIFF
--- a/src/routes/Player/Player.js
+++ b/src/routes/Player/Player.js
@@ -141,7 +141,7 @@ const Player = () => {
         }
     }, [platform.shell.active]);
     const castSubtitlesUrl = React.useRef(null);
-    const { castStarted, setSubtitles: setCastSubtitles } = useCastDevice(streamingServer.baseUrl ?? null);
+    const { castStarted, setSubtitles: setCastSubtitles } = useCastDevice();
     const onCastDeviceSelected = React.useCallback((deviceId) => {
         playOnDevice(deviceId, video.state.time);
         castStarted(deviceId);
--- a/src/routes/Player/Player.js
+++ b/src/routes/Player/Player.js
@@ -29,6 +29,7 @@ const { default: SideDrawerButton } = require('./SideDrawerButton');
 const { default: SideDrawer } = require('./SideDrawer');
 const usePlayer = require('./usePlayer');
 const { default: usePlayOnDevice } = require('./usePlayOnDevice');
+const { default: useCastDevice } = require('./useCastDevice');
 const { default: useKeyboardSeek } = require('./useKeyboardSeek');
 const useStatistics = require('./useStatistics');
 const useVideo = require('./useVideo');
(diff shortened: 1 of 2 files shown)


kKaskak:
hello this pr is incorrect because you are communicating with the server without using core? none of the issues you created should be together since these are distinct cases

1. subtitles need to be handled correctly when casting in either stremio-video or stremio-core 
2. casting device needs to be handled as separate pr
3.  leaving the player is a separate thing as well

please check again and use the correct approach to solve the issues in 3 diff prs instead

check contribution guide and how to md files use stremio-core stremio-video stremio-web 

each only if needed when ready open prs

dioKaratzas:
Closing this in favour of three focused PRs, one per issue, as requested — thanks for the direction. Will re-read the contribution guides in stremio-core / stremio-video / stremio-web first and open each only when it stands on its own.

kKaskak:
> @kKaskak you're right, thanks — server calls belong in core next to `PlayOnDevice`, not in the web layer. Reworked in [c608a2d](https://github.com/Stremio/stremio-web/commit/c608a2d51e14c4eff52ccb2554e5540c0abee791).
> 
> Core side is now [Stremio/stremio-core#1031](https://github.com/Stremio/stremio-core/pull/1031), adding `StreamingServer/StopOnDevice` and `StreamingServer/SetDeviceSubtitles` (plus a `StoppedOnDevice` event mirroring `PlayingOnDevice`). This PR just dispatches those actions through `core.transport`, so it needs the core PR merged and a `@stremio/stremio-core-web` release before it can go in — marking it draft until then.
> 
> The first part of the change (pausing local playback when a device is selected, instead of waiting for the `PlayingOnDevice` event — [Stremio/stremio-bugs#2731](https://github.com/Stremio/stremio-bugs/issues/2731)) is pure web and independent of that. Happy to split it into its own PR if you'd rather land it separately.



> @kKaskak you're right, thanks — server calls belong in core next to `PlayOnDevice`, not in the web layer. Reworked in [c608a2d](https://github.com/Stremio/stremio-web/commit/c608a2d51e14c4eff52ccb2554e5540c0abee791).
> 
> Core side is now [Stremio/stremio-core#1031](https://github.com/Stremio/stremio-core/pull/1031), adding `StreamingServer/StopOnDevice` and `StreamingServer/SetDeviceSubtitles` (plus a `StoppedOnDevice` event mirroring `PlayingOnDevice`). This PR just dispatches those actions through `core.transport`, so it needs the core PR merged and a `@stremio/stremio-core-web` release before it can go in — marking it draft until then.
> 
> The first part of the change (pausing local playback when a device is selected, instead of waiting for the `PlayingOnDevice` event — [Stremio/stremio-bugs#2731](https://github.com/Stremio/stremio-bugs/issues/2731)) is pure web and independent of that. Happy to split it into its own PR if you'd rather land it separately.

i dont like talking to claude here but ok :D remove it as co-author please 
split into 3 recheck core pr and remove excesive unnecessary comments

PR #1381, opened 2026-08-20
Player: show timestamp on seek bar hover

### LINKED ISSUE #1824: Timestamp when mousing over seek bar
### Stremio Platform

Windows

### Is your feature request related to a problem?

I don't like having to hold down the mouse button to see the time I'm seeking to as it hurts my hand to hold the mouse button down for more than a few seconds due to RSI.

### Describe the feature you'd like

I'd like for the timestamp to appear above the mouse cursor when mousing over the seek bar indicating exactly what time I'd be jumping to when I click.

### Use cases

When skipping intros for various reasons (weirdly loud theme song compared to the show that I don't feel like turning the volume down and then back up for, rapidly flashing lights in the intro, seen intro too many times and just want to get to the next episode, etc) this would make it easier for me to skip to just the right time without hurting my hand.

### Alternatives considered

Besides holding the mouse down and watching the time to the left of the bar change as I move the mouse, all I can think of for measuring out the amount of time I need to skip is by hitting the right arrow key enough times, but that's something I'd also like to avoid doing every episode due to hand pain.

### Mockups / Screenshots

https://i.imgur.com/hxkhB3J.png

### App Version

5.0.0-beta36

### Device Info

Desktop

### OS Version

Windows 11 Pro

### Additional context

_No response_

kKaskak:
Closes Stremio/stremio-features#1824

Hovering the seek bar now shows the timestamp you would seek to, above the cursor. Styled like the player menus, hidden while dragging since the seek label already shows the time there.

### DIFF
--- a/src/routes/Player/ControlBar/SeekBar/SeekBar.js
+++ b/src/routes/Player/ControlBar/SeekBar/SeekBar.js
@@ -1,6 +1,7 @@
 // Copyright (C) 2017-2023 Smart code 203358507
 
 const React = require('react');
+const ReactDOM = require('react-dom');
 const PropTypes = require('prop-types');
 const classnames = require('classnames');
 const debounce = require('lodash.debounce');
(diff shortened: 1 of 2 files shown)


Botsy (approved):
All looks good 👍

PR #1386, opened 2026-08-21
App: Add Touch press feedback

kKaskak:


Botsy (changes requested):
All looks good, but should we also add this effect to the icon buttons in meta details? Such as like / love / share / profile icon for non-premium users?

Botsy (approved):
Let's handle the icons in a separate PR, approving this one

PR #1393, opened 2026-08-21
Player: Improve SpeedMenu Slider selection & choices

dionjoshualobo:


kKaskak:
why do we need this? we laready redesigned the menu recently which allows for speeds from 0.25x to 4x which are standard speeds that are allowed by vlc as well more than 4x you cannot understand anything at all

dionjoshualobo:
We can make it so that the highest value possible is 4x. The purpose is more that it is just easier to type. My preferred speed (if I ever want to watch something faster) is 1.25x, which isn't an option in the buttons, and it is kind of hard to get that exact speed without a few over and unders with the slider. 

This just provides one other way for people to set the speed. Just type it and done.

dionjoshualobo:
In addition, VLC Player does let users to set precise speeds, but they have to do the extra work of going to the settings. 

mpv also lets you do it with its commands.

kKaskak:
> We can make it so that the highest value possible is 4x. The purpose is more that it is just easier to type. My preferred speed (if I ever want to watch something faster) is 1.25x, which isn't an option in the buttons, and it is kind of hard to get that exact speed without a few over and unders with the slider.
> 
> This just provides one other way for people to set the speed. Just type it and done.

its literally one single click, and you can choose with the buttons as well otherwise, i understand that the step is 0.1x not 0.01x but IMO simplicity is better here for the users. 

https://github.com/user-attachments/assets/a8c0e10d-ca26-4052-b084-477d266c5f42

kKaskak:
> > We can make it so that the highest value possible is 4x. The purpose is more that it is just easier to type. My preferred speed (if I ever want to watch something faster) is 1.25x, which isn't an option in the buttons, and it is kind of hard to get that exact speed without a few over and unders with the slider.
> > This just provides one other way for people to set the speed. Just type it and done.
> 
> its literally one single click, and you can choose with the buttons as well otherwise, i understand that the step is 0.1x not 0.01x but IMO simplicity is better here for the users.
> 
>  Screen.Recording.2026-08-22.at.20.19.13.mov


we need to reach middle ground here because overloading the menu with features or player UI isnt our goal

dionjoshualobo:
Yes, I do understand what you mean. But this feature won't add too much of a burden, it does not break any of the existing options, and it helps users choose the exact speed they want. 

In most other video players, 1.25 is a known option, which can't be achieved using buttons or the slider.

Well, at the end of the day you get to decide which feature stays and which feature goes. 

But since this is a feature I wanted, I will be using this locally. I thought there would be people who would want the same.

kKaskak:
> Yes, I do understand what you mean. But this feature won't add too much of a burden, it does not break any of the existing options, and it helps users choose the exact speed they want.
> 
> In most other video players, 1.25 is a known option, which can't be achieved using buttons or the slider.
> 
> Well, at the end of the day you get to decide which feature stays and which feature goes.
> 
> But since this is a feature I wanted, I will be using this locally. I thought there would be people who would want the same.

wdyt if we lower the slider step to 0.05 instead of 0.1 slider will still allow to do 0.25x to 4x same range as before

and redo the buttons to be

0.5x 1x 1.25x 1.5x 1.75x 2x 4x

kKaskak:
@dionjoshualobo

dionjoshualobo:
That should also be fine. Can I implement it?

kKaskak:
> That should also be fine. Can I implement it?

yeah you can in this pr

dionjoshualobo:
> yeah you can in this pr

I pushed the changes.

kKaskak (approved):

kKaskak:
> > yeah you can in this pr
> 
> I pushed the changes.

can you squash your commits now into one commit? thank you

PR #1395, opened 2026-08-24
Player: Fine tune subtitles delay controls

wjdob:
## Summary

- Change subtitle delay controls from 250ms to 100ms.
- Keep menu and `G`/`H` shortcut increments consistent.
- Use one `SUBTITLES_DELAY_STEP_MS` constant for millisecond logic and the stepper's seconds value.
- Use one directional snapping helper for menu and keyboard adjustments.
- Snap existing off-grid values only when the user adjusts them, so a 50ms remainder can return to zero or move onto the 100ms grid.

## Context

Related to [Stremio/stremio-features#1901](https://github.com/Stremio/stremio-features/issues/1901).

#1398 is stacked on this PR so its held-key path can use the same 100ms grid and snapping helper. This PR should be reviewed and merged first.

## Verification

- `npm run lint`
- `npm test -- --runInBand` (70 tests)
- Positive and negative grid arithmetic checks
- Local playback UAT passed with #1398

### DIFF
--- a/src/routes/Player/SubtitlesMenu/SubtitlesMenu.js
+++ b/src/routes/Player/SubtitlesMenu/SubtitlesMenu.js
@@ -118,7 +118,9 @@ const SubtitlesMenu = React.memo(React.forwardRef((props, ref) => {
         if (typeof props.selectedExtraSubtitlesTrackId === 'string') {
             if (props.extraSubtitlesDelay !== null && !isNaN(props.extraSubtitlesDelay)) {
                 if (typeof props.onExtraSubtitlesDelayChanged === 'function') {
-                    props.onExtraSubtitlesDelayChanged(Math.round(value * 1000));
+                    const delay = Math.round(value * 1000);
+                    const snappedDelay = delay > props.extraSubtitlesDelay ? Math.floor(delay / 100) * 100 : Math.ceil(delay / 100) * 100;
+                    props.onExtraSubtitlesDelayChanged(snappedDelay);
                 }
             }
         }
--- a/src/routes/Player/SubtitlesMenu/SubtitlesMenu.js
+++ b/src/routes/Player/SubtitlesMenu/SubtitlesMenu.js
@@ -10,6 +10,7 @@ const styles = require('./styles');
 const { t } = require('i18next');
 const { default: Stepper } = require('./Stepper');
 const { default: SubtitleVariant } = require('./SubtitleVariant');
+const { snapSubtitleDelay, SUBTITLES_DELAY_STEP_MS } = require('../subtitleDelay');
 
 const ORIGIN_PRIORITIES = [
     'LOCAL',
@@ -119,8 +120,7 @@ const SubtitlesMenu = React.memo(React.forwardRef((props, ref) => {
             if (props.extraSubtitlesDelay !== null && !isNaN(props.extraSubtitlesDelay)) {
                 if (typeof props.onExtraSubtitlesDelayChanged === 'function') {
                     const delay = Math.round(value * 1000);
-                    const snappedDelay = delay > props.extraSubtitlesDelay ? Math.floor(delay / 100) * 100 : Math.ceil(delay / 100) * 100;
-                    props.onExtraSubtitlesDelayChanged(snappedDelay);
+                    props.onExtraSubtitlesDelayChanged(snapSubtitleDelay(delay, delay - props.extraSubtitlesDelay));
                 }
             }
         }
@@ -216,7 +216,7 @@ const SubtitlesMenu = React.memo(React.forwardRef((props, ref) => {
                         label={'DELAY'}
                         value={props.extraSubtitlesDelay / 1000}
                         unit={'s'}
-                        step={0.1}
+                        step={SUBTITLES_DELAY_STEP_MS / 1000}
                         disabled={props.extraSubtitlesDelay === null}
                         onChange={onSubtitlesDelayChanged}
                     />
--- a/src/routes/Player/SubtitlesMenu/SubtitlesMenu.js
+++ b/src/routes/Player/SubtitlesMenu/SubtitlesMenu.js
@@ -118,7 +118,7 @@ const SubtitlesMenu = React.memo(React.forwardRef((props, ref) => {
         if (typeof props.selectedExtraSubtitlesTrackId === 'string') {
             if (props.extraSubtitlesDelay !== null && !isNaN(props.extraSubtitlesDelay)) {
                 if (typeof props.onExtraSubtitlesDelayChanged === 'function') {
-                    props.onExtraSubtitlesDelayChanged(value * 1000);
+                    props.onExtraSubtitlesDelayChanged(Math.round(value * 1000));
                 }
             }
         }
@@ -214,7 +214,7 @@ const SubtitlesMenu = React.memo(React.forwardRef((props, ref) => {
                         label={'DELAY'}
                         value={props.extraSubtitlesDelay / 1000}
                         unit={'s'}
-                        step={0.25}
+                        step={0.1}
                         disabled={props.extraSubtitlesDelay === null}
                         onChange={onSubtitlesDelayChanged}
                     />
--- a/src/routes/Player/subtitleDelay.ts
+++ b/src/routes/Player/subtitleDelay.ts
@@ -1,0 +1,11 @@
+const SUBTITLES_DELAY_STEP_MS = 100;
+
+const snapSubtitleDelay = (delay: number, direction: number) => {
+    const snap = direction > 0 ? Math.floor : Math.ceil;
+    return snap(delay / SUBTITLES_DELAY_STEP_MS) * SUBTITLES_DELAY_STEP_MS;
+};
+
+export {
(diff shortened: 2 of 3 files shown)


src/routes/Player/SubtitlesMenu/SubtitlesMenu.js
wjdob, line 1: JavaScript floating-point arithmetic can represent repeated 0.1 steps as values like 0.7999999999. Multiplying by 1000 could then produce 799.9999, which downstream truncation might turn into 799 ms. Math.round(value * 1000) guarantees the intended whole-millisecond value, such as exactly 800 ms.

wjdob (commented):
Adjustment of 250ms default subtitle delay step to 100ms in support of [https://github.com/Stremio/stremio-features/issues/1901](https://github.com/Stremio/stremio-features/issues/1901)

kKaskak:
hey there, i personally did not see any of the users prefer lower ms step for subtitles usually 200-300ms is the perfect step, can you elaborate more in which cases except your addon this will be better UI UX for the users? Since currently what i see is just longer waiting times to reach the correct offset + more clicks or more hold time since you can hold to speed it up.

kKaskak:
@wjdob ?

wjdob:
1. Anyone with OCD - in my testing with a Game of Thrones S01E01 episode, I could not match audio to subtitle using the default 250ms step. My unofficial seen/heard cue subtitle sync aligned them correctly. Test was done using the same subtitle file - so it is not a wrong subtitle framerate edge-case issue that may result in a drift.
2. The subtitle step issue affects users using addons such as Strelingo - you can land on a good dual language subtitle pair, but the default 250ms step prevents proper sync, rendering the good dual language subtitle pair useless if primary subtitle does not match release/timing perfectly.
3. In most of the cases with popular movies/series - you can find a perfectly aligned subtitle anyways. The manual subtitle delay adjustment covers the edge cases where there is no properly matched subtitle to the release, but you can use the extra precision to correct sync. 100ms offers greater precision control.
4. In terms of the UI/UX: the added time of "fast-stepping" by decreasing 250ms to 100ms (2.5x longer on paper) can be remediated by adjusting how fast is the stepping mechanism when holding the +/- steps (increase the speed of the existing fast-forward stepping mechanism by 2.5x). The net result is the same time to arrive at the ballpark delay time, with extra couple seconds spent fine-tuning using the 100ms stepping.
5. I am also now aware that under donate version of Stremio the Live Sync feature is available - the live sync solution takes care of the syncing much better than manually stepping anyway. So I would argue that:
a. 100ms gives better precision control to users who do not use donate version.
b. if increased precision is too cumbersome, user can move to donate version with live sync which is a much more convenient option anyway
6. After double-checking how VLC does it - VLC uses 50ms by default unless customized in config. So I am being generous with 100ms

kKaskak:
could you implement hold to speed up in a separate pr then? the longer you hold the faster it gets?

wjdob:
One follow-up from testing #1395 together with #1398:

1. Combining 100ms fine-stepping with the existing 250ms held step can otherwise leave a 50ms remainder - for example, `2.25s`.
2. From that point, adding or subtracting a fixed 100ms keeps the remainder forever: `2.15s`, `2.05s`, `1.95s`, etc. This means the user cannot arrive at `0s` using the fine-step controls.
3. I addressed this by snapping the next manual adjustment in the direction selected by the user. For example:
   - `+0.05s` and minus becomes `0s`
   - `+0.05s` and plus becomes `+0.1s`
   - `-0.05s` and plus becomes `0s`
   - `-0.05s` and minus becomes `-0.1s`
4. The current delay is not changed automatically when loading the subtitle. Snapping only occurs when the user presses plus/minus or uses `G`/`H`, so the action still follows the direction requested by the user.
5. I kept this adjustment in #1395 because that PR defines the 100ms precision grid. #1398 can remain an independent change for held-step acceleration.

I tested positive and negative values in the combined local build, including crossing zero, and the controls no longer leave the user stuck on the 50ms remainder.

src/routes/Player/SubtitlesMenu/SubtitlesMenu.js
```diff
@@ -118,7 +118,9 @@ const SubtitlesMenu = React.memo(React.forwardRef((props, ref) => {
         if (typeof props.selectedExtraSubtitlesTrackId === 'string') {
             if (props.extraSubtitlesDelay !== null && !isNaN(props.extraSubtitlesDelay)) {
                 if (typeof props.onExtraSubtitlesDelayChanged === 'function') {
-                    props.onExtraSubtitlesDelayChanged(value * 1000);
+                    const delay = Math.round(value * 1000);
+                    const snappedDelay = delay > props.extraSubtitlesDelay ? Math.floor(delay / 100) * 100 : Math.ceil(delay / 100) * 100;
```
Botsy: This snaps by comparing against the previous delay, while `useSubtitles.ts` snaps by
baking the direction into floor-vs-ceil. Same rule, two formulations — they match
today but will drift. Suggest a shared `snapDelay(current, direction)` used by both.

src/routes/Player/SubtitlesMenu/SubtitlesMenu.js
```diff
@@ -214,7 +216,7 @@ const SubtitlesMenu = React.memo(React.forwardRef((props, ref) => {
                         label={'DELAY'}
                         value={props.extraSubtitlesDelay / 1000}
                         unit={'s'}
-                        step={0.25}
+                        step={0.1}
```
Botsy: `step={0.1}` is in seconds while the snap grid two hops away is in ms. Changing one
without the other silently produces values the snapper immediately rounds off. Worth
deriving this from a shared `SUBTITLES_DELAY_STEP_MS / 1000`.

src/routes/Player/useSubtitles.ts
```diff
@@ -282,11 +282,11 @@ const useSubtitles = ({
     }, [streamStateChanged, video]);
 
     const increaseDelay = useCallback(() => {
-        changeDelay((video.state.extraSubtitlesDelay ?? 0) + 250);
+        changeDelay(Math.floor((video.state.extraSubtitlesDelay ?? 0) / 100) * 100 + 100);
```
Botsy: Correct for negatives (verified −150 → −100 up, −200 down). Same note as the menu
handler though: this is the second implementation of one rule. Extracting it also
gives the accelerated held-key path somewhere to snap through.

Botsy (changes requested):
I reviewed and tested locally. The arithmetic is correct - I ran both snap paths
over on-grid, off-grid and negative values and all looks good on UI side 👍 

Two things I'd like resolved before merge, both structural rather than behavioural:

1. The same snapping rule is implemented twice in two different formulations (menu
   compares against the previous delay, keyboard bakes direction into floor-vs-ceil).
   They agree today, but nothing enforces that. Worth extracting a single
   `snapDelay(current, direction)` helper.

2. `100` is a magic number in three places, and `step={0.1}` in the JSX is silently
   coupled to it across a unit boundary (seconds vs ms). A single `SUBTITLES_DELAY_STEP_MS` constant with the stepper deriving `/1000` would make that coupling explicit.

Heads up on a conflict: #(hold-acceleration) introduces `SUBTITLES_DELAY_REPEAT_STEP_MS = 250` with an accelerated step of 250/500/750/1000, and its accelerated keyboard path calls `changeDelay(current + direction * step)` with no snapping at all. Under this PR's grid the odd multiples land off-grid, so held-key repeats would drift off the same grid that tap-repeats snap onto. Whichever lands second needs the base constant unified to 100 and the accelerated path routed through the shared helper - which is another argument for extracting it here.

wjdob:
I addressed the requested changes:

1. The menu and keyboard paths now use the same `snapSubtitleDelay` helper instead of implementing the rule separately.
2. The 100ms value now comes from one `SUBTITLES_DELAY_STEP_MS` constant. The stepper derives its seconds value from that constant, so the two units cannot drift independently.
3. The helper accepts the proposed delay and direction. This preserves the user's requested direction for positive, negative and off-grid values.
4. #1398 is now stacked on this branch. Its held keyboard path routes through the same helper, while the 250ms held travel remains separate from the 100ms precision grid.
5. I reran lint, all 70 tests and the production build. Combined local playback UAT with #1398 passed.

src/routes/Player/subtitleDelay.ts
```diff
@@ -0,0 +1,11 @@
+const SUBTITLES_DELAY_STEP_MS = 100;
```
Botsy, line 3: Copyright header is missing. Once you add it, I will approve the PR 👍

Botsy (changes requested):

wjdob:
I addressed the latest review and refreshed the branch:

1. Added the repository copyright header to `subtitleDelay.ts`.
2. Rebased on current `development`, including the upstream change that preserves subtitle delay when switching external tracks.
3. Re-ran lint, all 70 tests, the production build and `git diff --check`. All pass; the build retains the two existing bundle-size warnings.

@Botsy, requesting re-review.

Botsy (approved):
All looks good 👍

kKaskak (approved):

PR #1396, opened 2026-08-25
feat(player): add cue-based subtitle sync

wjdob:
## Summary

- Add a compact Live Sync control for external subtitles.
- Let users mark `Audio heard` and `Subtitle seen` in either order, then apply the calculated constant offset automatically.
- Add `Shift+H` and `Shift+J` shortcuts, a reset action, and clear captured-marker styling.
- Preserve the existing subtitle delay path and stream-state persistence.

## Behavior

- Capture markers only during active playback with an available external subtitle.
- Use monotonic timestamps and account for playback speed.
- Cancel incomplete marker pairs on pause, buffering, seek, speed change, stream change, subtitle change, or manual delay adjustment.
- Prevent held cue-sync shortcuts from repeating.
- Keep the existing subtitle menu layout and present both actions as centered two-line controls.

## Scope

This corrects constant offset for external subtitles. It does not change embedded-subtitle timing or correct FPS drift, edited cuts, or piecewise timing differences.

## Translation dependency

Depends on [Stremio/stremio-translations#1092](https://github.com/Stremio/stremio-translations/pull/1092).

The feature was developed and tested against the local translation checkout. No dependency on the contributor fork is committed. The translation PR should merge first; this branch will then pin its official upstream commit before being marked ready for review. Until then, a clean CI checkout may report the three new translation keys as missing.

## Verification

- Dedicated translation-key scan
- `npm run lint`
- `npm test -- --runInBand` (70 tests)
- Manual UAT on Stremio Desktop for external-subtitle synchronization and the revised controls

## Related requests

- [Stremio/stremio-features#1901](https://github.com/Stremio/stremio-features/issues/1901)
- [Stremio/stremio-features#750](https://github.com/Stremio/stremio-features/issues/750)
- [Stremio/stremio-features#914](https://github.com/Stremio/stremio-features/issues/914)

### DIFF
--- a/src/common/Shortcuts/Shortcuts.tsx
+++ b/src/common/Shortcuts/Shortcuts.tsx
@@ -47,7 +47,9 @@ const ShortcutsProvider = ({ children, onShortcut }: Props) => {
             lastRepeatTime.current.set(repeatKey, now);
         }
 
-        SHORTCUTS.forEach(({ name, combos }) => combos.forEach((keys) => {
+        SHORTCUTS.forEach(({ name, combos, repeat = true }) => combos.forEach((keys) => {
+            if (event.repeat && !repeat) return;
+
             const modifers = (keys.includes('Ctrl') === ctrlKey)
                 && (keys.includes('Shift') === shiftKey)
                 && !altKey
--- a/src/common/Shortcuts/shortcuts.json
+++ b/src/common/Shortcuts/shortcuts.json
@@ -75,6 +75,18 @@
                 "combos": [["G"], ["H"]]
             },
             {
+                "name": "subtitlesSyncAudio",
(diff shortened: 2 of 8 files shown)


wjdob:
UAT:

Screenshot 2026-08-24 210849

kKaskak:
hey thank you so much for your PR however we already have this feature implemented called Live Sync. This is basically the same just with buttons if i understood correctly.
check https://stremio.com/donate
here is a preview as well:
Screenshot 2026-08-25 at 12 50 35

kKaskak:
however if you are developing an addon which will support both langs or show one of the langs same way we do the logic is in stremio-video so you can check it out thank you again

PR #1398, opened 2026-08-26
Player: accelerate held subtitle delay adjustments

wjdob:
## Summary

- Keep the repeat interval fixed and increase only the subtitle-delay step while the control is held.
- Use the same hold delay, repeat interval and elapsed-time multiplier for the on-screen controls and the `G` and `H` shortcuts.
- Keep a normal click or key press at one 100ms step.
- Snap off-grid values in the selected direction so an existing 50ms remainder can still reach zero.
- Apply no extra step when a held on-screen control is released.
- Cancel keyboard holds on key release, focus loss or opening a player menu.
- Leave subtitle size, vertical position and visual layout unchanged.

## Hold behavior

The first repeat starts after 250ms. Repeats remain fixed at 100ms.

| Hold duration | Delay step |
| --- | ---: |
| Under 1 second | 300ms |
| 1-2 seconds | 400ms |
| 2-3 seconds | 500ms |
| 3 seconds or longer | 600ms |

## Context

Follow-up to the hold-acceleration request in [#1395](https://github.com/Stremio/stremio-web/pull/1395#issuecomment-5423913628).

Related to [Stremio/stremio-features#1901](https://github.com/Stremio/stremio-features/issues/1901).

## Verification

- `npm run lint`
- `npm test` (70 tests)
- `git diff --check`
- Local playback UAT passed with the merged #1395 behavior

### DIFF
--- a/src/common/Shortcuts/Shortcuts.tsx
+++ b/src/common/Shortcuts/Shortcuts.tsx
@@ -5,7 +5,7 @@ import shortcuts from './shortcuts.json';
 const SHORTCUTS = shortcuts.map(({ shortcuts }) => shortcuts).flat();
 
 export type ShortcutName = string;
-export type ShortcutListener = (combo: number, key: string) => void;
+export type ShortcutListener = (combo: number, key: string, repeat: boolean) => void;
 
 interface ShortcutsContext {
     grouped: ShortcutGroup[],
@@ -60,7 +60,7 @@ const ShortcutsProvider = ({ children, onShortcut }: Props) => {
 
             if (modifers && keyMatched) {
                 const combo = combos.indexOf(keys);
-                listeners.current.get(name)?.forEach((listener) => listener(combo, key));
+                listeners.current.get(name)?.forEach((listener) => listener(combo, key, repeat));
 
                 onShortcut(name as ShortcutName, combo, key);
             }
--- a/src/routes/Player/SubtitlesMenu/Stepper/Stepper.tsx
+++ b/src/routes/Player/SubtitlesMenu/Stepper/Stepper.tsx
@@ -4,6 +4,11 @@ import classNames from 'classnames';
 import Icon from '@stremio/stremio-icons/react';
 import { Button } from 'stremio/components';
 import { useInterval, useTimeout } from 'stremio/common';
+import {
+    getSubtitleDelayStepMultiplier,
+    SUBTITLES_DELAY_REPEAT_DELAY_MS,
+    SUBTITLES_DELAY_REPEAT_INTERVAL_MS,
+} from '../../subtitleDelay';
 import styles from './Stepper.less';
 
 const clamp = (value: number, min?: number, max?: number) => {
@@ -21,16 +26,19 @@ type Props = {
     min?: number,
     max?: number,
     disabled?: boolean,
+    accelerate?: boolean,
     onChange: (value: number) => void,
 };
 
-const Stepper = ({ className, label, value, unit, step, min, max, disabled, onChange }: Props) => {
+const Stepper = ({ className, label, value, unit, step, min, max, disabled, accelerate = false, onChange }: Props) => {
     const { t } = useTranslation();
 
     const localValue = useRef(value);
+    const holdStartedAt = useRef(0);
+    const hasRepeated = useRef(false);
 
-    const interval = useInterval(100);
-    const timeout = useTimeout(250);
+    const interval = useInterval(SUBTITLES_DELAY_REPEAT_INTERVAL_MS);
+    const timeout = useTimeout(SUBTITLES_DELAY_REPEAT_DELAY_MS);
 
     const cancel = () => {
         interval.cancel();
@@ -51,27 +59,45 @@ const Stepper = ({ className, label, value, unit, step, min, max, disabled, onCh
 
     const updateValue = useCallback((delta: number) => {
         onChange(clamp(localValue.current + delta, min, max));
-    }, [onChange]);
+    }, [max, min, onChange]);
+
+    const startRepeating = useCallback((direction: number) => {
+        cancel();
+        holdStartedAt.current = performance.now();
+        hasRepeated.current = false;
+        let repeatValue = localValue.current;
+
+        timeout.start(() => interval.start(() => {
+            if (!accelerate) {
+                updateValue(direction * step);
+                return;
+            }
+
+            hasRepeated.current = true;
+            const heldFor = performance.now() - holdStartedAt.current;
+            const multiplier = getSubtitleDelayStepMultiplier(heldFor);
+            repeatValue = clamp(repeatValue + direction * step * multiplier, min, max);
+            onChange(repeatValue);
+        }));
+    }, [accelerate, max, min, onChange, step, updateValue]);
 
     const onDecrementMouseDown = useCallback(() => {
-        cancel();
-        timeout.start(() => interval.start(() => updateValue(-step)));
-    }, [updateValue]);
+        startRepeating(-1);
+    }, [startRepeating]);
 
     const onDecrementMouseUp = useCallback(() => {
         cancel();
-        updateValue(-step);
-    }, [updateValue]);
+        if (!accelerate || !hasRepeated.current) updateValue(-step);
+    }, [accelerate, step, updateValue]);
 
     const onIncrementMouseDown = useCallback(() => {
-        cancel();
-        timeout.start(() => interval.start(() => updateValue(step)));
-    }, [updateValue]);
+        startRepeating(1);
+    }, [startRepeating]);
 
     const onIncrementMouseUp = useCallback(() => {
         cancel();
-        updateValue(step);
-    }, [updateValue]);
+        if (!accelerate || !hasRepeated.current) updateValue(step);
+    }, [accelerate, step, updateValue]);
 
     useEffect(() => {
         localValue.current = value;
--- a/src/routes/Player/SubtitlesMenu/SubtitlesMenu.js
+++ b/src/routes/Player/SubtitlesMenu/SubtitlesMenu.js
@@ -212,6 +212,7 @@ const SubtitlesMenu = React.memo(React.forwardRef((props, ref) => {
                 <div className={styles['settings-header']}>{t('PLAYER_SUBTITLES_SETTINGS')}</div>
                 <div className={styles['settings-list']}>
                     <Stepper
+                        accelerate
                         className={styles['stepper']}
                         label={'DELAY'}
                         value={props.extraSubtitlesDelay / 1000}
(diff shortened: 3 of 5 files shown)


src/common/Shortcuts/Shortcuts.tsx
```diff
@@ -33,19 +33,24 @@ const isInputFocused = () => {
 const ShortcutsProvider = ({ children, onShortcut }: Props) => {
     const listeners = useRef<Map<ShortcutName, Set<ShortcutListener>>>(new Map());
     const lastRepeatTime = useRef<Map<string, number>>(new Map());
+    const keyDownAt = useRef<Map<string, number>>(new Map());
```
Botsy: `keyDownAt` entries are written but never removed — there's no `keyup` or `blur` handler clearing them. Please, add one.

src/common/Shortcuts/Shortcuts.tsx
```diff
@@ -33,19 +33,24 @@ const isInputFocused = () => {
...
+        if (!repeat) {
+            keyDownAt.current.set(repeatKey, now);
+        }
         if (repeat) {
-            const now = Date.now();
             const last = lastRepeatTime.current.get(repeatKey) ?? 0;
             if (now - last < REPEAT_THROTTLE_MS) return;
             lastRepeatTime.current.set(repeatKey, now);
         }
+        const heldForMs = repeat ? now - (keyDownAt.current.get(repeatKey) ?? now) : 0;
```
Botsy: `keyDownAt` (and `lastRepeatTime`) are written but never cleared — there's no `keyup` or `blur` handler. Normal press/release is fine, since the next non-repeat keydown overwrites the entry.

The gap is a repeat keydown arriving without a preceding non-repeat one, which happens when the window loses and regains focus with the key still held. `heldForMs` is then measured against a timestamp from whenever the key was last pressed — potentially minutes ago — and saturates at the top tier on the very first repeat, jumping the delay by a full second immediately. The `?? now` fallback covers a *missing* entry, not a stale one.

Clearing the entry on `keyup`, and on window `blur` since keyup is lost once focus leaves, would close both paths.

src/routes/Player/SubtitlesMenu/Stepper/Stepper.tsx
Botsy: The step gets bigger the longer you hold, but the repeats keep firing at the same speed — every 100 ms. So both go up at once.

After 3 seconds of holding, each repeat is worth 1 second of delay, and there are 10 of them per second. That's 10 seconds of delay for every 1 second you hold the button. Before this change it was 2.5 seconds.

In practice: hold the button for 3.5 seconds and the delay moves about 19 seconds. The same press used to move 8 seconds.

| You hold for | Delay moves |
|---|---|
| 1 s | 1.75 s |
| 2 s | 6.75 s |
| 3 s | 14.25 s |
| 3.5 s | 19.25 s |

Since subtitle delay is usually adjusted by a fraction of a second, this makes it hard to land on the value you want. You shoot past it, hold the other button to come back, and shoot past it again.

What I'd suggest: as the step grows, make the repeats fire further apart. Then holding longer still gets you there faster, but not 4× faster. A `getAcceleratedInterval` next to `getAcceleratedStep` would do it.

Just don't slow the repeats down too much — if you fully cancel out the bigger steps, every tier moves at the same speed and it won't feel like it's accelerating at all. Letting it reach about 3× the normal speed keeps that feeling and puts a 3.5 second hold at around 14 seconds.

Botsy (changes requested):

wjdob:
I addressed the requested changes:

1. Shortcut state is cleared on `keyup` and window blur. A repeat event without a current keydown is ignored, so restoring focus cannot reuse an old hold duration and jump directly to the maximum tier.
2. The held step still increases over time, but repeat intervals now increase from 100ms to 175ms. This reduces an ideal 3.5-second mouse hold from approximately 19.5 seconds to approximately 14.1 seconds in the combined build.
3. `G` and `H` use the same elapsed-time interval schedule rather than relying only on the operating system's keyboard repeat cadence.
4. Held movement keeps an internal ideal value and emits on the active tap grid. This preserves the average 250ms coarse travel without leaving the user stuck on a 50ms remainder.
5. This PR is stacked on #1395 so the held keyboard path uses its shared 100ms snapping helper. #1395 should merge first.
6. I reran lint, all 70 tests and the production build. Combined local playback UAT with #1395 passed.

src/routes/Player/SubtitlesMenu/Stepper/Stepper.tsx
```diff
@@ -53,25 +61,44 @@ const Stepper = ({ className, label, value, unit, step, min, max, disabled, onCh
...
+        const repeat = () => {
+            const heldFor = performance.now() - holdStartedAt.current;
+            const delta = accelerate ? getAcceleratedStep(repeatStep, heldFor) : repeatStep;
+            const interval = accelerate ?
+                getAcceleratedInterval(SUBTITLES_DELAY_REPEAT_INTERVAL_MS, heldFor)
+                :
+                SUBTITLES_DELAY_REPEAT_INTERVAL_MS;
+            const snap = direction > 0 ? Math.floor : Math.ceil;
+            repeatValue += direction * delta;
+            onChange(clamp(snap(repeatValue / step) * step, min, max));
```
Botsy: Minor: snapping in float seconds gives slightly different values than the keyboard path, which snaps in whole milliseconds via `snapSubtitleDelay` (`0.6 / 0.1` is `5.999999999999999`, so `Math.floor` drops a cell). Not visible in practice — the value still moves every tick and always lands on the grid — but reusing the helper here (passing `step` in as an argument) would remove both the duplication and the float edge.

Botsy (changes requested):
Please, add missing copyright headers in files.

wjdob:
I addressed the latest review and refreshed the branch:

1. Added the repository copyright header to `subtitleDelayAcceleration.ts`. The stacked `subtitleDelay.ts` receives the same header through #1395.
2. The Stepper no longer performs a separate floating-point floor/ceil calculation in seconds. It rounds the repeated value and active step to whole milliseconds, passes both through the shared `snapSubtitleDelay` helper, then converts the result back to seconds.
3. This keeps values such as `0.6s` on the correct `0.6s` cell instead of floating-point division snapping them to `0.5s`, while preserving directional snapping and the held ideal value.
4. Rebased on current `development` and the updated #1395, removing the previous merge commit from this branch.
5. Re-ran lint, all 70 tests, the production build, focused snapping checks and `git diff --check`. All pass; the build retains the two existing bundle-size warnings.

@Botsy, requesting re-review.

src/routes/Player/SubtitlesMenu/Stepper/Stepper.tsx
```diff
@@ -53,25 +62,48 @@ const Stepper = ({ className, label, value, unit, step, min, max, disabled, onCh
...
+
+    const onDecrementMouseDown = useCallback(() => {
+        startRepeating(-1);
+    }, [startRepeating]);
 
     const onDecrementMouseUp = useCallback(() => {
         cancel();
         updateValue(-step);
-    }, [updateValue]);
+    }, [step, updateValue]);
```
Botsy, line 100: The repeat loop keeps its own counter (`repeatValue`) instead of reading `localValue.current`, which suggests the `value` prop can fall behind during a fast hold. But mouseup goes back to `localValue.current` via `updateValue`, so the release doesn't get that protection.

Say the delay starts at 0 and you hold minus for 3.5 s. The loop has told the parent `-14.0`, and the repeat before that was `-13.0` (each step is 1 s at the top tier). On release, mouseup emits `localValue.current - 0.1`:

- if the prop caught up, that's `-14.1` — one extra 100 ms step past where you released
- if the prop is one repeat behind, that's `-13.1` — the delay jumps **backwards 0.9 s** the moment you let go

Simplest fix: only apply the mouseup step when no repeat fired, since a plain click is the only case that needs it. Or compute it from `repeatValue` so the release uses the same counter the loop trusts.

Botsy (changes requested):

src/routes/Player/SubtitlesMenu/Stepper/Stepper.tsx
wjdob: Agreed with the proposed change. I reproduced both release cases locally before applying it:

1. The repeat loop finished at `-14.0s`.
2. With the prop caught up, the previous mouseup path emitted `-14.1s`.
3. With the prop one repeat behind, it emitted `-13.1s`.
4. I applied the simpler suggested fix: record whether a repeat fired and apply the mouseup step only when it did not. A normal click still applies one 100ms step, while release after a hold keeps the last repeated value.
5. I tested both directions, every acceleration tier, current and delayed parent values, and the non-accelerated size and position steppers. The 68-case release matrix, lint, all 70 tests and the production build pass.

The branch is also rebased on current `development` now that #1395 is merged. @Botsy, requesting re-review.

kKaskak (changes requested):
I think this has become more complicated than it needs to be. The goal is simple: the longer you hold the button, the faster the value changes, up to a reasonable cap.

Can we keep the repeat interval fixed and just increase the step size as the hold continues? If it gets too fast, lower the maximum step instead of adding another schedule to slow the repeats down.

Using multiples of the existing 100ms step should also remove the extra bookkeeping needed to reconcile 250ms hold steps with 100ms taps.

wjdob:
I redesigned this following the proposed direction:

1. The repeat interval now remains fixed at 100ms.
2. Held adjustments use multiples of the existing 100ms step. The multiplier starts at 3x, increases once per second and is capped at 6x.
3. The on-screen plus/minus controls and the `G`/`H` shortcuts now use the same hold delay, repeat interval and multiplier schedule.
4. I removed the variable interval schedule and the separate 250ms coarse-step reconciliation.
5. Normal taps remain 100ms, off-grid values still snap in the selected direction and releasing a held control does not add another step.
6. I reran lint, all 70 tests, the production build and `git diff --check`. Manual playback UAT passed.
7. The previous implementation remains available on the [snapshot branch](https://github.com/wjdob/stremio-web/tree/snapshot/pr-1398-dual-schedule) for reference.

@kKaskak, @Botsy, requesting re-review.

wjdob:
I rebased this branch on current `development` and resolved the upstream subscription changes:

1. Kept upstream's new `useFileDropListener` implementation.
2. Migrated the subtitle-delay hold handler from the removed `onShortcut` helper to `useShortcut`.
3. Preserved the keyboard repeat flag used to distinguish a tap from a held `G`/`H` adjustment.
4. Kept upstream's stable video action callbacks and the fixed-interval acceleration behavior unchanged.
5. `pnpm lint`, all 70 tests, the production build and `git diff --check` pass. The production build retains the two existing bundle-size warnings.

@kKaskak, @Botsy, requesting re-review after the rebase.

kKaskak (approved):

PR #1404, opened 2026-08-30
Discover: Add Catalog preview in a mobile bottom sheet

mrcanelas:
## Summary
On phones (width `<= 1000px`), tapping a catalog item in Discover opens `MetaPreview` in the existing `BottomSheet` instead of navigating away. This applies in both portrait and landscape. Desktop still uses the sidebar preview.

Portrait keeps a vertical layout, with the Watch/Show action as icon plus label. Landscape follows MetaDetails: content on the left, a column of round action buttons on the right.

The sheet reuses the Calendar bottom sheet (overlay drag handle, optional title hidden for this preview, enter/exit animation, flush layout). Swipe-to-dismiss works on the content, like a native sheet: if the content cannot scroll, or it is scrolled to the top, pulling down starts the close. Watch/Show closes the sheet and opens MetaDetails.

## Why
The Discover sidebar preview is hidden below 1000px, so a tap previously left Discover. A sheet keeps browsing in place on phones. Landscape still has room for a sheet if the content and actions are split the way MetaDetails already does.

## Test plan
- [ ] Portrait, width `<= 1000px`: tap a poster opens the sheet and does not change the route
- [ ] Landscape, width `<= 1000px`: tap a poster opens the sheet (does not navigate); content on the left, round action buttons stacked on the right; sheet width matches Calendar (~90%, not full bleed)
- [ ] Watch/Show closes the sheet and navigates to MetaDetails
- [ ] Library, watched, and trailer actions work inside the sheet
- [ ] Backdrop tap, swipe down on content (at the top of the scroll), and Escape close the sheet with the exit animation
- [ ] With the sheet open, open catalog filters: Escape closes only the filters modal; a second Escape closes the sheet
- [ ] Desktop: first tap selects the sidebar preview, second tap on the same item navigates
- [ ] Calendar day details: sheet still opens and closes; rotation still dismisses it; purple glow and heading/content gap match the previous Calendar sheet


Botsy (changes requested):
Tested locally against the test plan and these are my findings and suggestions for improvement:
### Issues

**Landscape: tap navigates to MetaDetails** — ❌ the gate is width-only, so landscape phones get the sheet, and its layout is broken (handle overlaps content, buttons overflow). Since landscape should navigate, add the orientation condition:

    const isPortrait = useMediaQuery('(orientation: portrait)');
    const useMobilePreview = isMobile && isPortrait;

and delete the new landscape CSS in `MetaPreview/styles.less`, `Discover/styles.less` and `ActionButton/styles.less` rather than fixing it - it also overlaps the existing `@phone-landscape` block.

**Escape isn't stacking-aware** — ❌ the new `document` keydown closes the sheet regardless of what's above it, so sheet + filters modal both close. Compare with the top-most check in ModalDialog.js:45; rendering through `src/router/Modal` gets that plus focus lock, which the sheet currently lacks despite `aria-modal="true"`.

**Calendar sheet regressions** — ❌ rotation no longer closes it (the `useOrientation` effect was dropped in the rewrite), and `box-shadow`/`gap` were removed from `.container` unconditionally so Calendar loses its glow and its spacing shifts. Scope those to `&:not(.flush)`.

### Suggestions

- **Fold `useOrientation` into the new hook.** The PR adds a second mechanism for the same question, so the older one can just delegate — which also drops its resize listener and its bogus `[window.innerWidth, window.innerHeight]` effect deps:

      const useOrientation = (): DeviceOrientation =>
          useMediaQuery('(orientation: portrait)') ? 'portrait' : 'landscape';

  Handy here too, since the landscape fix above needs `(orientation: portrait)` anyway.

- `useMediaQuery.ts` — the `typeof window` guard is dead code in a browser-only bundle (the effect below dereferences `window` unguarded), and the `onChange()` on :16 is a redundant render the lazy initializer already covers.

- Discover.js:39 — `1000px` hardcoded in JS while the CSS hides the sidebar at `@xsmall`; exporting it from Less keeps one source of truth. `useMobilePreview` on :40 is a bare alias of `isMobile`.

- BottomSheet.tsx:67-74 — five state flags and two competing close paths (`onTransitionEnd` *and* a timeout) for a two-state animation, mostly because `onClose` doubles as "please close" and "closed". One phase machine plus an `onCloseRequest` prop would cover it; I'd keep the timeout as the only completion signal since `transitionend` doesn't fire under `prefers-reduced-motion`.

- BottomSheet.tsx:10 — `ANIMATION_DURATION` duplicates the `0.25s` in the stylesheet; drive one from the other.

- BottomSheet.tsx:40-52 — `isScrolledToTop` calls `getComputedStyle` on every `touchmove`; resolve the scrollable ancestor once at touchstart.

- The new `:global(.wide)` / `padding: 0 !important` overrides reach into ActionButton from MetaPreview's stylesheet; a variant prop would be sturdier than overriding from three levels up.

- Backdrop `aria-label={'Close'}` is untranslated (`t('BUTTON_CLOSE')` exists), and `aria-label={title}` leaves the Discover sheet unlabelled since it passes no title.

- Discover.js:122 — the new `Number.isInteger` guard swallows the click when `data-index` is missing, where the link used to navigate.

mrcanelas:
@Botsy, thanks for the review; I went through the issues and suggestions.

On landscape: that one is intentional. Earlier we had the sheet portrait-only (landscape went to MetaDetails), but @kKaskak  asked to keep the preview in landscape as well, with the MetaDetails split, content on the left, a column of round icon actions on the right. Watch/Show still closes the sheet and opens MetaDetails. The test plan you reviewed still described the old portrait-only behaviour, the description is updated now, including the extra checks around stacked Escape, Calendar glow/rotation, and landscape width matching the Calendar sheet (~90%).

The rest of the findings should be addressed:

- Escape is stacking-aware (same top-most modal check as `ModalDialog`) and the sheet renders through `Modal`, so it also gets focus lock
- Calendar: rotation closes again; `gap` / `box-shadow` are scoped to `&:not(.flush)`; the iOS `clip-path` is only on the Discover flush sheet so the Calendar glow isn’t clipped
- `useOrientation` now just delegates to `useMediaQuery('(orientation: portrait)')`
- `useMediaQuery` uses `useSyncExternalStore` (no dead typeof window guard / extra mount render)
- JS breakpoint comes from `@xsmall` via Less `:export`, the `useMobilePreview` alias is gone
- BottomSheet is a small phase machine (`idle` -> `entering` -> `entered` -> `exiting`) with `onCloseRequest`; the timeout is the only close-completion signal
- `ANIMATION_DURATION` is driven from the Less value
- Scrollable ancestor is resolved once on `touchstart`
- ActionButton has an `icon` / `wide` variant instead of `:global(.wide)` / `!important`
- Backdrop uses `t('BUTTON_CLOSE')`; Discover passes `ariaLabel` from the item name
- Missing `data-index` no longer swallows the click, the link can navigate

Happy to retest whenever you have a moment.

PR #1408, opened 2026-09-02
Player: Persist video scale state

kKaskak:


Botsy (dismissed):
Tested and all works as expected 👍

Botsy (approved):

PR #1413, opened 2026-09-03
App: Confirm streaming server URL from search params

esp4ce:
### Problem

`?streamingServerUrl=` is applied as soon as the app sees it, without any validation or user interaction:

```js
// src/App/SearchParamsHandler.js
if (streamingServerUrl) {
    core.transport.dispatch({ action: 'Ctx', args: { action: 'UpdateSettings', args: { ...profile.settings, streamingServerUrl } } });
    core.transport.dispatch({ action: 'Ctx', args: { action: 'AddServerUrl', args: streamingServerUrl } });
}
```

Any page that gets the browser to open a Stremio link can repoint the streaming server of the running app, and the value is not even checked to be an `http(s)` URL. Adding a server from Settings already requires an explicit user action — this was the only path that did not.

### Change

The parameter is now ignored unless it parses as a URL with an `http:` or `https:` scheme, using the same `new URL()` check as `useStreamingServerUrls` plus a scheme allowlist. When it is valid, a dialog shows the URL and `UpdateSettings` / `AddServerUrl` are dispatched only on confirm. Cancel, backdrop click and Escape leave the settings untouched.

The dialog portals into `document.body` like `ShortcutsModal` and `GamepadModal`; `ModalDialog` is not usable here because `SearchParamsHandler` sits outside the per-route `ModalsContainerProvider`. The URL is displayed with `direction: ltr; unicode-bidi: isolate-override` so a bidi override character in the parameter cannot reorder what the user reads.

Needs Stremio/stremio-translations#1104 for `SETTINGS_SERVER_CONFIGURE_CONFIRM`. The title reuses `SETTINGS_SERVER_CONFIGURE_TITLE`, the buttons `BUTTON_CANCEL` and `BUTTON_CONFIRM`.

An invalid parameter is dropped silently rather than raising an error toast, to avoid adding hardcoded text — happy to add a key for it instead.

### DIFF
--- a/src/App/SearchParamsHandler.js
+++ b/src/App/SearchParamsHandler.js
@@ -4,6 +4,16 @@ const React = require('react');
 const { deepEqual } = require('fast-equals');
 const { useCore } = require('stremio/core');
 const { withCoreSuspender, useProfile, useToast } = require('stremio/common');
+const { default: StreamingServerUrlModal } = require('./StreamingServerUrlModal');
+
+const isValidStreamingServerUrl = (url) => {
+    try {
+        const { protocol } = new URL(url);
+        return protocol === 'http:' || protocol === 'https:';
+    } catch (_) {
+        return false;
+    }
+};
 
 const SearchParamsHandler = () => {
     const core = useCore();
@@ -11,6 +21,7 @@ const SearchParamsHandler = () => {
     const toast = useToast();
 
     const [searchParams, setSearchParams] = React.useState({});
+    const [requestedStreamingServerUrl, setRequestedStreamingServerUrl] = React.useState(null);
 
     const onLocationChange = () => {
         const { origin, hash, search } = window.location;
@@ -22,32 +33,41 @@ const SearchParamsHandler = () => {
         });
     };
 
+    const onStreamingServerUrlConfirm = React.useCallback(() => {
+        core.transport.dispatch({
+            action: 'Ctx',
+            args: {
+                action: 'UpdateSettings',
+                args: {
+                    ...profile.settings,
+                    streamingServerUrl: requestedStreamingServerUrl,
+                },
+            },
+        });
+        core.transport.dispatch({
+            action: 'Ctx',
+            args: {
+                action: 'AddServerUrl',
+                args: requestedStreamingServerUrl,
+            },
+        });
+        toast.show({
+            type: 'success',
+            title: `Using streaming server at ${requestedStreamingServerUrl}`,
+            timeout: 4000,
+        });
+        setRequestedStreamingServerUrl(null);
+    }, [requestedStreamingServerUrl, profile.settings]);
+
+    const onStreamingServerUrlCancel = React.useCallback(() => {
+        setRequestedStreamingServerUrl(null);
+    }, []);
+
     React.useEffect(() => {
         const { streamingServerUrl } = searchParams;
 
-        if (streamingServerUrl) {
-            core.transport.dispatch({
-                action: 'Ctx',
-                args: {
-                    action: 'UpdateSettings',
-                    args: {
-                        ...profile.settings,
-                        streamingServerUrl,
-                    },
-                },
-            });
-            core.transport.dispatch({
-                action: 'Ctx',
-                args: {
-                    action: 'AddServerUrl',
-                    args: streamingServerUrl,
-                },
-            });
-            toast.show({
-                type: 'success',
-                title: `Using streaming server at ${streamingServerUrl}`,
-                timeout: 4000,
-            });
+        if (streamingServerUrl && isValidStreamingServerUrl(streamingServerUrl)) {
+            setRequestedStreamingServerUrl(streamingServerUrl);
         }
     }, [searchParams]);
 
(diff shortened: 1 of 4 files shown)


kKaskak:
hey there is a toast which displays the value of the connection string to the user as soon as the user connects to a url which contains the server url in the params? whats the reason behind adding a confirmation prompt?

esp4ce:
yeah the toast shows up after the fact, UpdateSettings and AddServerUrl are already dispatched by then and it's gone in 4s. so it tells you what happened, it doesn't give you a chance to say no. miss it and your server url is just changed and added to the saved list.

the part that bugged me is that the param doesn't have to come from the user, any page can send the browser to #/?streamingServerUrl=... and there's no validation on the value either, any string goes through right now.

that said if the one click link is the main use case i'm happy to cut the friction. eg only prompt when the url isn't already in ctx.streamingServerUrls, so you'd get the dialog once and never again for your own server. or keep the toast and put an undo on it instead of a modal. whatever you prefer, i'll rework it

kKaskak:
> yeah the toast shows up after the fact, UpdateSettings and AddServerUrl are already dispatched by then and it's gone in 4s. so it tells you what happened, it doesn't give you a chance to say no. miss it and your server url is just changed and added to the saved list.
> 
> the part that bugged me is that the param doesn't have to come from the user, any page can send the browser to #/?streamingServerUrl=... and there's no validation on the value either, any string goes through right now.
> 
> that said if the one click link is the main use case i'm happy to cut the friction. eg only prompt when the url isn't already in ctx.streamingServerUrls, so you'd get the dialog once and never again for your own server. or keep the toast and put an undo on it instead of a modal. whatever you prefer, i'll rework it
> 

Isn't it validated for url?
Even if any value goes through what's the security / vulnerability implication here?

esp4ce:
`url::Url` parses stuff that isn't a server. `?streamingServerUrl=javascript:x` passes, then the core panics building the settings request off it and the ui goes blank till reload. so any page linking to `#/?streamingServerUrl=...` can crash the app. the http/https allowlist fixes that.

the crash isn't the main thing though. once applied, the core does `GET <url>/settings` and trusts the response, if it has `remoteHttps` it then calls `/get-https?authKey=<account auth key>&...` against that same url. so a server the user never picked gets handed the account authKey. that's why i wanted a real confirm before it applies, the toast only tells you after.

it's account level so i'd rather not post the full steps here, happy to send them privately if you've got a security contact.

and if the one click link matters i can only prompt when the url isn't already in `ctx.streamingServerUrls`. lmk.

kKaskak:
> `url::Url` parses stuff that isn't a server. `?streamingServerUrl=javascript:x` passes, then the core panics building the settings request off it and the ui goes blank till reload. so any page linking to `#/?streamingServerUrl=...` can crash the app. the http/https allowlist fixes that.
> 
> the crash isn't the main thing though. once applied, the core does `GET <url>/settings` and trusts the response, if it has `remoteHttps` it then calls `/get-https?authKey=<account auth key>&...` against that same url. so a server the user never picked gets handed the account authKey. that's why i wanted a real confirm before it applies, the toast only tells you after.
> 
> it's account level so i'd rather not post the full steps here, happy to send them privately if you've got a security contact.
> 
> and if the one click link matters i can only prompt when the url isn't already in `ctx.streamingServerUrls`. lmk.

right ok, no need i will open next stacked pr for all cases: 

URL state  - Behaviour:
```
   Invalid or unsupported scheme  - Reject
   Already selected  -  Do nothing
   Already in ctx.streamingServerUrls - Switch immediately; toast only if selection changed
   Default 127.0.0.1:11470 / localhost:11470  - Silent
   New loopback URL -  Apply and show toast
   New non-loopback URL - Confirm once, then save it
```

esp4ce:
matrix looks right, one thing i'd push back on: "new loopback url — apply + toast, no confirm".

localhost isn't really a trust boundary here. once a url is selected the core does GET <url>/settings and takes the response at face value, and if that response has a non-empty remoteHttps and you're logged in it follows up with GET <url>/get-https?authKey=<account authKey>&ipAddress=... against the same url (update_remote_url -> get_https_endpoint in stremio-core/src/models/streaming_server.rs). so any page can do #/?streamingServerUrl=http://127.0.0.1:<port> and whatever is listening on that port gets the account authKey in a query string. no confirm = no chance to stop it.

so i'd either confirm for new loopback urls too, or better, gate the get-https call so the authKey only ever goes to a server the user explicitly added, regardless of host. the second option fixes it in core and doesn't cost the one-click ux anything.

it's account level so i didn't want to put the full repro in the thread, still happy to send it over if you've got a security contact, it's a couple of lines.

separate from the matrix: ?streamingServerUrl=javascript:x (or anything url::Url parses that isn't a base url) makes the core panic when it builds the settings request, page goes blank till reload. the http/https allowlist in this pr covers it from the web side but it should probably be handled in core too.

on this pr, do you want me to extend 1413 to cover the whole matrix, or close it and fold it into your stacked one? either works, just don't want to double up.

kKaskak:
> any page can do #/?streamingServerUrl=http://127.0.0.1: and whatever is listening on that port gets the account authKey in a query string. no confirm = no chance to stop it.

That would require a very specific flow tho, its an extremely unlikely case don't you think?

1. A malicious or compromised local application listens on an arbitrary port.
2. A page navigates Stremio to ?streamingServerUrl=http://127.0.0.1:<that-port>.
3. The local listener returns a forged /settings response containing remoteHttps.
4. Core sends /get-https?authKey=... to that listener.

esp4ce:
step 1 isn't "malware", any localhost port that echoes attacker-influenced json works - dev server, electron app, extension bridge. low bar.

not high severity, agreed. but cheap to kill: only send get-https to a url already in ctx.streamingServerUrls. loopback stays silent, no dialog. i can do that in core.

extend 1413 for the matrix or fold into your stacked pr?

kKaskak:
keeping loopback silent, the listener has to be attacker-controlled anyway. 1414 supersedes 1413, will credit you in the merge.

PR #1414, opened 2026-09-03
App: Confirm only untrusted streaming server URLs

kKaskak:
credit to @esp4ce for opening an issue for this


Botsy (approved):
All looks good 👍

PR #1419, opened 2026-09-04
Load visible catalogs by model index and cancel queued measurements

kKaskak:
Search and Board used filtered DOM positions as model indexes and left queued measurements alive. Share row mapping and debounce ownership while preserving route activity and Board preload.

### DIFF
--- a/src/common/useVisibleCatalogs.ts
+++ b/src/common/useVisibleCatalogs.ts
@@ -1,0 +1,47 @@
+// Copyright (C) 2017-2026 Smart code 203358507
+
+import { useCallback, useLayoutEffect, useMemo, useRef } from 'react';
+import getVisibleChildrenRange from './getVisibleChildrenRange';
+import { useRouteActive } from './useRouteFocused';
+
+const debounce: (callback: () => void, wait: number) => { (): void, cancel: () => void } = require('lodash.debounce');
+
+type Props = {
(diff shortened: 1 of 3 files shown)


src/common/useVisibleCatalogs.ts
```diff
@@ -0,0 +1,47 @@
...
+        .filter(({ catalog }) => catalog.content?.type !== 'Err' || catalog.content.content !== 'EmptyContent'), [catalogs]);
+
+    const measure = useCallback(() => {
+        if (!active || !scrollContainerRef.current || catalogRows.length === 0) return;
+
+        const range = getVisibleChildrenRange(scrollContainerRef.current);
+        if (range === null) return;
+
+        const firstRow = catalogRows[Math.max(0, range.start - leadingRows)];
+        const lastIndex = range.end - leadingRows;
```
Botsy: Unbounded index - `catalogRows[lastIndex]` is `undefined` the moment this container gets a trailing child, and the throw happens in the layout effect (line 40) with no error boundary anywhere in the app → white screen. Worth clamping:

    const maxRow = catalogRows.length - 1;
    const lastIndex = Math.min(maxRow, range.end - leadingRows);

(same for `firstRow` on line 29)

Botsy (changes requested):

Botsy (approved):

PR #1423, opened 2026-09-04
Restore temporary playback speed when a hold is canceled

kKaskak:
Opening a menu during temporary 2x playback lost the speed restoration. Give keyboard and pointer holds one cancellation owner, including focus loss and teardown.


Botsy (approved):
tested and looks good 👍

PR #1424, opened 2026-09-04
Derive next video popup visibility from the current stream

kKaskak:
The next-video popup could remain open after the next item disappeared. Derive visibility from current playback data and keep dismissal tied to the stream.

### DIFF
--- a/src/routes/Player/Player.js
+++ b/src/routes/Player/Player.js
@@ -104,7 +104,12 @@ const Player = () => {
     const [speedMenuOpen, , closeSpeedMenu, toggleSpeedMenu] = useBinaryState(false);
     const [statisticsMenuOpen, openStatisticsMenu, closeStatisticsMenu, toggleStatisticsMenu] = useBinaryState(false);
     const [castDevicesMenuOpen, , closeCastDevicesMenu, toggleCastDevicesMenu] = useBinaryState(false);
-    const [nextVideoPopupOpen, openNextVideoPopup, closeNextVideoPopup] = useBinaryState(false);
+    const [nextVideoPopupDismissal, setNextVideoPopupDismissal] = React.useState(null);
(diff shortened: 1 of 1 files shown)


src/routes/Player/Player.js
```diff
@@ -104,7 +104,12 @@ const Player = () => {
     const [speedMenuOpen, , closeSpeedMenu, toggleSpeedMenu] = useBinaryState(false);
     const [statisticsMenuOpen, openStatisticsMenu, closeStatisticsMenu, toggleStatisticsMenu] = useBinaryState(false);
     const [castDevicesMenuOpen, , closeCastDevicesMenu, toggleCastDevicesMenu] = useBinaryState(false);
-    const [nextVideoPopupOpen, openNextVideoPopup, closeNextVideoPopup] = useBinaryState(false);
+    const [nextVideoPopupDismissal, setNextVideoPopupDismissal] = React.useState(null);
+    const nextVideoPopupOpen = player.nextVideo !== null &&
+        nextVideoPopupDismissal?.stream !== video.state.stream &&
```
Botsy: `nextVideoPopupDismissal?.stream` is `undefined` when nothing has been dismissed, so this reads `undefined !== video.state.stream`. Works today only because `useVideo` initializes `stream: null` — if the video implementation ever emits `propChanged('stream', undefined)`, "never dismissed" and "dismissed" become the same thing and the popup stops showing for the rest of the session.

```suggestion
        (nextVideoPopupDismissal === null || nextVideoPopupDismissal.stream !== video.state.stream) &&
```

Also worth a note here that dismissal is scoped to stream *identity* — same assumption `persistedVideoScale` makes on line 600, but it's easier to miss inside a render expression.

Botsy (changes requested):

Botsy (approved):

PR #1432, opened 2026-09-04
Reset image failure state at the source boundary

kKaskak:
Image reset failure state through a layout effect after a source change. Reset the inner image owner at the source boundary, preserving fallback and retry behavior.

### DIFF
--- a/src/components/Image/Image.tsx
+++ b/src/components/Image/Image.tsx
@@ -1,6 +1,6 @@
 // Copyright (C) 2017-2023 Smart code 203358507
 
-import React, { useCallback, useLayoutEffect, useState } from 'react';
+import React, { useCallback, useState } from 'react';
 
 type Props = {
     className: string,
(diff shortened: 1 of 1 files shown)


src/components/Image/Image.tsx
```diff
@@ -11,7 +11,7 @@ type Props = {
     onError: (event: React.SyntheticEvent<HTMLImageElement>) => void,
 };
```
Botsy: Please, extend the image props here, and make the props optional. Possible solution:
```
type Props = React.ImgHTMLAttributes<HTMLImageElement> & {
    className?: string,
    src?: string,
    alt?: string,
    fallbackSrc?: string,
    renderFallback?: () => React.ReactNode,
    onError?: (event: React.SyntheticEvent<HTMLImageElement>) => void,
};
```

Botsy (changes requested):

Botsy (dismissed):

PR #1439, opened 2026-09-04
Enable React Hooks lint checks

kKaskak:
Add React Hooks lint coverage and name wrapper components correctly. Hook-order violations fail lint; dependency diagnostics remain warnings while the existing backlog is reviewed.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -2,6 +2,7 @@ import globals from 'globals';
 import pluginJs from '@eslint/js';
 import tseslint from 'typescript-eslint';
 import pluginReact from 'eslint-plugin-react';
+import pluginReactHooks from 'eslint-plugin-react-hooks';
 import stylistic from '@stylistic/eslint-plugin';
 
 export default [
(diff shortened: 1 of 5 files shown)


Botsy (dismissed):

eslint.config.mjs
```diff
@@ -45,6 +47,8 @@ export default [
     },
     {
         rules: {
+            'react-hooks/rules-of-hooks': 'error',
+            'react-hooks/exhaustive-deps': 'warn',
```
Botsy, line 51: `warn` is the right call — I ran this config and there are **287** existing violations, so `error` would block the repo. But a warning count that large means nobody would notice violation **288**, and the description's "while the existing backlog is reviewed" has no mechanism behind it.

One line makes it a ratchet instead of a suggestion (`package.json`, line 13):

```json
"lint": "eslint src --max-warnings 287",
```

Count can then only go down. For reference, it's concentrated in a handful of files — `Player.js` 48, `usePlayerOptions.ts` 18, `Slider.js` 12, `Stream.js` 10, `usePlayer.js` 10 — so it's tractable to chip away at.

PR #1441, opened 2026-09-05
Shell: Open magnets and torrent files

kKaskak:


src/common/useTorrent.js
```diff
@@ -14,45 +14,51 @@ const useTorrent = () => {
...
-    }, []);
+    }, [createTorrent]);
     React.useEffect(() => {
         if (streamingServer.torrent !== null) {
             const [, { type }] = streamingServer.torrent;
-            if (type === 'Ready') {
+            if (type === 'Ready' || type === 'Err') {
                 clearTimeout(createTorrentTimeout.current);
                 toast.remove(parsingToastId.current);
             }
```
Botsy, line 54: Adding `Err` here fixes the 20s wait, but it also removes the error message.

Previously an `Err` left the timeout running, so the user eventually got
"Failed to parse magnet link." Now the loading toast is just dismissed and
nothing takes its place — a torrent rejected by the streaming server fails
silently.

`DeepLinkHandler` only handles `Ready`, so nothing downstream surfaces it either.

Could we show the error toast here instead? Something like:

    if (type === 'Ready') {
        clearTimeout(createTorrentTimeout.current);
        toast.remove(parsingToastId.current);
    } else if (type === 'Err') {
        clearTimeout(createTorrentTimeout.current);
        toast.remove(parsingToastId.current);
        toast.show({ type: 'error', title: 'Failed to open torrent.', timeout: 8000 });
    }

The `Err` variant carries `content`, so a more specific message is possible too.

TRtomasz (dismissed):
works on windows

PR #1442, opened 2026-09-05
Fix navigation history and direct-link back actions

kKaskak:


Botsy (approved):

Botsy:
@kKaskak tested and all looks good, but please fix those useEffect warnings.

PR #1445, opened 2026-09-06
Player: Preserve audio language across episodes

### LINKED ISSUE #2682: Audio Track and Subtitles
### Stremio Platform

Windows (Beta)

### Describe the bug

It seems that the Preferred Audio Track and Preferred Subtitle Track settings don't work as expected.

What I'd like is for Stremio to remember the last manually selected audio and subtitle tracks and automatically keep using them when the next episode starts.

For example, if I select the Japanese audio track and English subtitles (or disable subtitles), those choices should remain selected for every following episode, as long as those tracks are available. Right now, I have to manually select them again every episode.

This would make watching TV series much more seamless, especially for anime and multilingual content.

**I've also confirmed that this issue occurs on the Android mobile version, so it doesn't appear to be limited to the desktop app.**

### To Reproduce

Play a TV series with multiple audio and subtitle tracks.
Manually select a different audio track (e.g. Japanese) and a subtitle track (e.g. English, or set subtitles to Off).
Finish the episode or skip to the next one.
Notice that the selected audio and/or subtitle tracks are not preserved and must be selected again manually.

### Expected behavior

**Expected behavior:**

Keep the selected audio track when moving to the next episode.
Keep the selected subtitle track (or "Off") when moving to the next episode.
Only fall back to another track if the previously selected one isn't available.

### Screenshots

_No response_

### App Version

v5.0.23

### Device Info

Redmi note 13 Pro

### OS Version

Windows 11 Pro

### Player Info

Exo, VLC, MPV

### Browser Info

_No response_

### Logs

```shell

```

### Notes

_No response_

kKaskak:


TRtomasz (dismissed):
looks ok on windows

PR #1446, opened 2026-09-06
Fix cast playback handoff, subtitles and cleanup

### LINKED ISSUE #2731: [Bug]: Casting to a device keeps playing locally as well (v5 desktop)
### Stremio Platform

macOS (Stremio 5 beta, ARM build)

### Describe the bug

When a cast device is selected from the player, playback starts on the device but the local playback keeps running, so the same stream plays twice (local audio/video plus the device).

The desktop cast path goes through the streaming server (`StreamingServer` / `PlayOnDevice` in stremio-core). In `stremio-web`, `Player.js` only reacts *afterwards*, when the core emits the `PlayingOnDevice` event:

```js
const onCoreEvent = (name) => {
    if (name === 'PlayingOnDevice') {
        playingOnExternalDevice.current = true;
        onPauseRequested();
    }
};
```

Until that event arrives (server has to reach the device, open the session and answer), nothing pauses the local player, so both play. `onCastDeviceSelected` itself does not touch local playback at all:

```js
const onCastDeviceSelected = React.useCallback((deviceId) => {
    playOnDevice(deviceId, video.state.time);
    closeCastDevicesMenu();
}, [playOnDevice, video.state.time]);
```

### To Reproduce

1. Start any stream in the player (desktop shell, streaming server running).
2. Open the cast menu and select a Chromecast / DLNA TV.
3. Playback starts on the device while the local playback continues.

### Expected behavior

Selecting a cast device should stop (pause) local playback immediately, so the stream only plays on the device.

### Version

5.0.0-beta.39 (also reproduced on earlier 5 betas)

### Additional context

Proposed fix: pause the local playback in `onCastDeviceSelected` instead of waiting for the `PlayingOnDevice` event. PR: https://github.com/Stremio/stremio-web/pull/1376

### LINKED ISSUE #2732: [Bug]: Subtitles are not sent to the cast device (v5 desktop)
### Stremio Platform

macOS (Stremio 5 beta, ARM build)

### Describe the bug

While casting to a Chromecast / DLNA device, subtitles are rendered on the local screen only. The device shows no subtitles at all.

The streaming server does support subtitles for cast sessions. `GET /casting/{deviceId}/player` returns, on a live session:

```json
{"volume":1,"time":406192.1,"paused":false,"state":3,"length":3162460,
 "source":"http://127.0.0.1:11470/<hash>/0",
 "subtitlesSrc":"https://subs5.strem.io/en/download/subencoding-stremio-utf8/src-api/file/...",
 "subtitlesDelay":0,"subtitlesSize":2}
```

and the devices themselves advertise the capability (`"playerUIRoles":["playpause","seek","dub","subtitles","volume"]`).

Stremio 4.4 used this: it wrapped the endpoint in the `stremio-cast` client and set `subtitlesSrc` on the remote player. In `stremio-web` (v5) the only thing ever sent to the device is the initial `PlayOnDevice` action with `device`, `source` and `time` — `subtitlesSrc` is never set, so the device has no subtitles.

### To Reproduce

1. Start a stream and select a subtitles track.
2. Cast to a Chromecast or DLNA TV.
3. Subtitles show locally, but not on the device.

### Expected behavior

The selected subtitles track should be pushed to the cast device (and updated when the track is changed mid-playback), as in v4.

### Version

5.0.0-beta.39

### Additional context

Proposed fix: POST `{"subtitlesSrc": "<track url>"}` to `/casting/{deviceId}/player` when casting starts and whenever the selected track changes. PR: https://github.com/Stremio/stremio-web/pull/1376

### LINKED ISSUE #2733: [Bug]: Leaving the player does not stop playback on the cast device (v5 desktop)
### Stremio Platform

macOS (Stremio 5 beta, ARM build)

### Describe the bug

While casting to a device, pressing Escape (or otherwise leaving the player) closes the local player, but the device keeps playing the stream. There is no way to stop it from the app afterwards — the session stays alive on the device until it is stopped from the device itself.

`stremio-web` never sends anything to end the cast session. The only cast-related action dispatched is `StreamingServer` / `PlayOnDevice` (see `usePlayOnDevice.ts`); there is no counterpart action, and the player's unmount path does not touch the device.

The streaming server does expose a stop: the `/casting/{deviceId}/player` endpoint implements the stremio-cast protocol, where `POST {"source": null}` stops the playback (this is exactly what `stremio-cast`'s `Client.stop()` did in v4).

### To Reproduce

1. Start a stream and cast it to a Chromecast / DLNA TV.
2. Press Escape / navigate back to leave the player.
3. The local player closes, the device keeps playing; `GET /casting/{deviceId}/player` still reports `"state":3`.

### Expected behavior

Leaving the player should stop playback on the cast device.

### Version

5.0.0-beta.39

### Additional context

Proposed fix: POST `{"source": null}` to `/casting/{deviceId}/player` when the player unmounts. PR: https://github.com/Stremio/stremio-web/pull/1376

### LINKED ISSUE #2789: Casting always restarts the video at 0, and the position the UI sends is discarded
**Stremio version:** 5.1.27 macOS (Apple silicon). Streaming server 4.21.0 (bundled), node 16.20.2.
**Devices:** LG 42LM760S-ZB (DLNA), and the same code path applies to Chromecast.

**What happens**
Start an episode in the app, watch to 6:11, then cast. The device starts at 00:00. The same
happens after every subtitle or audio-track change, because each one reloads the stream.

**The UI does its part**
The desktop UI posts the position. Captured on the server:

```
POST /casting/<devID>/player  body={"source":"http://localhost:11470/<infohash>/7?","time":369000}
```

That matches `usePlayOnDevice.ts` in stremio-web and `play_on_device` in stremio-core, which
send `{source, time}`.

**Where it is lost**
Two places in the streaming server, both in `server.js`:

1. `DLNAClient.prototype.play` (and `ChromecastClient.prototype.play`) take only `srcURL` and
   then do `this.mediaStatus.time = 0`. The dispatch in `Player.prototype.middleware` passes
   just `params.source`, so `time` never reaches `play()`:

   ```js
   isset(params.source) && (params.source ? args[method = "play"] = params.source : method = "close")
   ```

2. Even after passing it through, it is overwritten before use. `play()` awaits
   `castingUtils.getVideoInfo` (an ffmpeg probe of the torrent stream, seconds on a cold
   torrent) and only calls `delayedPlayFromStatus()` afterwards. During that window the
   still-playing previous stream keeps emitting UPnP position events, and
   `_updateStatusField("time", …)` overwrites `mediaStatus.time`. Observed concretely:
   requesting 6:11 landed the device at 18:11, the position of the previous stream.

**Fix**
Pass the position to `play()` and let it survive the probe:

```js
// dispatch
args[method = "play"] = [ params.source, params.time ]
// DLNAClient.play(srcURL, startAt) / ChromecastClient.play(srcURL, startAt)
this.mediaStatus.time = parseInt(startAt, 10) || 0
// capture before the probe, restore after it
var self = this, wantedAt = this.mediaStatus.time;
… self.mediaStatus.length = 1e3 * info.duration, self.mediaStatus.time = wantedAt, self.delayedPlayFromStatus();
```

Chromecast additionally needs `this.seekTime`, since `playFromStatus` builds its content id
from that rather than from `mediaStatus.time`.

With both in place, casting at 6:11 produces `ffmpeg -ss 371 …` and the device starts there.

**A second, related defect in the same area**
Once positions from the renderer are processed at all, they are double counted:

```js
this.mediaStatus[field] = this.seekTime + 1e3 * parseInt(value, 10);
```

The cast stream is produced with `-copyts`, so the renderer already reports absolute time.
After resuming at 14:37 the app showed 29:29 within seconds. Guarding it covers renderers
that do report from zero:

```js
this.mediaStatus[field] = (function (t, s) { return t >= s ? t : s + t; })(1e3 * parseInt(value, 10), this.seekTime || 0);
```

Related: #2731, #2732, #2733 and PR Stremio/stremio-web#1446 cover casting handoff and
subtitles, but not the start position. Patch script and evidence:
https://github.com/WouterP0lman/stremio-macos-cast-patches (patches 11, 12 and 13).

kKaskak:
Fixes Stremio/stremio-bugs#2731
Fixes Stremio/stremio-bugs#2732
Fixes Stremio/stremio-bugs#2733


kKaskak:
tested on chromecast by @TRtomasz

PR #1454, opened 2026-09-14
Check movie sources before leaving Discover

cristiantorresf19191919:
When users select movies with no sources, they repeatedly leave Discover and land on an empty stream screen. This draft adds an opt-in check before that navigation. The dialog stays in Discover, distinguishes returned sources from empty responses and addon failures, and offers View sources, Check again, Manage addons, or an explicit View details anyway action.

The implementation checks one user-selected movie. It does not yet filter the whole catalog, cache results across dialogs, or verify playback compatibility. The broader proposal remains documented separately from the implemented scope. Keyboard focus enters the dialog and returns to the movie card on dismissal; slow checks stop waiting after 15 seconds without claiming that the movie is unavailable.

Dependencies:
- Core: https://github.com/Stremio/stremio-core/pull/1056
- Translations: https://github.com/Stremio/stremio-translations/pull/1116

This draft needs a core package release before it can be marked ready. The published core dependency does not yet expose source_preview; the review guide explains how to build and link the companion branch locally. The translation dependency is temporarily pinned to the reviewed fork commit and must move upstream after merge. An older core produces an explanatory fallback instead of crashing.

Validation: web lint passes with existing warnings, all 70 existing tests pass, the translation scan passes through the installed Jest, and both the production web build and combined local WASM/web build pass. In-browser fixture checks covered positive, empty, HTTP 503, external-only, timeout, source navigation, Escape, Enter, Tab, and focus restoration. The fixture server recorded no video downloads. Desktop and 390 × 844 layouts were inspected. A standalone type check reports the same 37 errors on unchanged development, with none in the new dialog. Core and translation checks are recorded in their linked drafts.

See docs/source-check-review.md for scope, local setup, source paths, and screenshots. The screenshots use fictional titles and controlled addon responses.

Source check in Discover


kKaskak:
hello, im sorry to disappoint but this pr does not implement any feature, or fixes a bug that we could accept, this is the standard app behaviour to enter from discover page to meta details page to see the streams. also the only case where there are no streams is when the movie is in cinemas only which is a rare case, we had a feature before which showed a cinema label but we removed it due inaccuracy, also streams are provided from the addon and their place is logically in the details page not in discover page, this breaks the app in so many diff ways

kKaskak:
please add more addons so you will never see the no streams message, except when there is literally no streams available due to unreleased or in cinema only (when in cinema only you almost always have streams as well )

PR #1460, opened 2026-09-20
Add manual Skip Intro control

ThiaJay:
This is the Web side of the Skip Intro work from Stremio/stremio-core#1064, following the architecture discussed in Stremio/stremio-core#1063.

It reads the `introOutro.intro` timing supplied by Core and shows a manual Skip Intro button only while playback is within that intro section.

The button uses Stremio’s existing seek path, is not shown for live playback and does not make any provider requests from the Web UI itself.

It reuses the existing `PLAYER_SKIP_INTRO` translation key.

The Web build, tests and lint all passed. I also tested the Web change against the final Core prebuild and the combined Core plus Web integration passed.

### Reference image

Genuine running-build capture from the integrated Core + Web preview. The image contains only Stremio page content.

Skip Intro running in Stremio


kKaskak:
rthis is already implemented feature check stremio.com/plans

